PR #4, opened 2022-11-07
i577 Fix bug when creating new work

bkiahstroud:
Ref https://gitlab.com/notch8/palni-palci/-/issues/577

When creating a new Work, if you don't make any changes on the Sharing tab, no permissions params are sent. Short-circuit if permissions params aren't present


labradford (approved):
lgtm

alishaevn (approved):

PR #308, opened 2022-11-17
actions use latest tag

bkiahstroud:


bkiahstroud:
The actions v0.0.4 tag has issues. Because pals is already using a tagged version of the actions, we should be fine just closing this without making any changes for now

Issue #149, opened 2022-11-08
Add appearance option to update favicon


orangewolf:
# Summary
Pals would like to give their tenants the option for a favicon icon. This work was done for BL: the option is in Setting -> Appearance.

Related BL ticket: https://gitlab.com/notch8/britishlibrary/-/issues/90

# Acceptance Criteria

- [ ] Favicon icon option for each tenant is in place

# Screenshots or Video

# Testing Instructions

# Notes

Confused how to fill this out? Check out the [playbook](https://playbook-staging.notch8.com/en/notch8/process/issues)

DiemBTran:
Linked items:

- Epic: Backlog Tickets

labradford:
`

[2022-11-30T20:37:24.559339 #1] INFO -- : [b1593af86e553ce31e441418ca9d14c0] Sending event e14b5692b63b4b068cd64aab854b5c2c to Sentry
--
Wed, Nov 30 2022 12:37:24 pm | F, [2022-11-30T20:37:24.681273 #1] FATAL -- : [b1593af86e553ce31e441418ca9d14c0]
Wed, Nov 30 2022 12:37:24 pm | F, [2022-11-30T20:37:24.681541 #1] FATAL -- : [b1593af86e553ce31e441418ca9d14c0] NoMethodError (undefined method `favicon_will_change!' for #<Site:0x00007f7ada37b930>
Wed, Nov 30 2022 12:37:24 pm | Did you mean? favicon_cache):
Wed, Nov 30 2022 12:37:24 pm | F, [2022-11-30T20:37:24.684820 #1] FATAL -- : [b1593af86e553ce31e441418ca9d14c0]
Wed, Nov 30 2022 12:37:24 pm | F, [2022-11-30T20:37:24.685159 #1] FATAL -- : [b1593af86e553ce31e441418ca9d14c0] app/forms/hyrax/forms/admin/appearance.rb:355:in `update!'
Wed, Nov 30 2022 12:37:24 pm | [b1593af86e553ce31e441418ca9d14c0] app/controllers/hyrax/admin/appearances_controller.rb:28:in `update'

`

DiemBTran:
**This passes SoftServ QA:** See attached screen recording.

tested on:
- https://qa-13.commons-archive.org/admin/appearance?locale=en#favicon
[...]

PR #313, opened 2022-11-22
add favicon uploads to appearance dashboard

labradford:
# Summary
ref: #149 
add ability upload custom favicon per tenant

# Screenshots / Video

# Expected Behavior
User can upload favicon in appearance panel
favicon will appear in browser tab

# Notes
Code from British Library courtesy of @orangewolf and @alishaevn


app/controllers/hyrax/homepage_controller.rb
```diff
@@ -63,6 +63,8 @@ def index
       end
     end
 
+    def browserconfig; end
```
bkiahstroud, line 66: What is this for?

app/uploaders/favicon_uploader.rb
```diff
@@ -0,0 +1,23 @@
...
+  # 196×196	favicon-196.png Chrome for Android home screen icon
+  # rubocop:enable Style/AsciiComments
+
+  [32, 57, 76, 96, 128, 192, 228, 196, 120, 152, 180].each do |i|
+    version "v#{i}".to_sym do
+      process resize_to_limit: [i, i]
+    end
+  end
+
+  def extension_whitelist
```
bkiahstroud, line 20: Does this method need to have this name, or can we rename it to avoid using "whitelist"?

app/controllers/hyrax/homepage_controller.rb
bkiahstroud: Is it related to this file? `app/views/hyrax/homepage/browserconfig.xml.erb`

spec/controllers/sites_controller_spec.rb
```diff
@@ -52,6 +52,23 @@
...
+        FaviconUploader.enable_processing = false
+        file = fixture_file_upload('/images/favicon.png', 'image/png')
+        Site.instance.update(favicon: file)
+        FaviconUploader.enable_processing = true
+      end
+
+      it "#update with remove_favicon deletes a favicon" do
+        expect(Site.instance.favicon?).to be true
+        post :update, params: { id: Site.instance.id, remove_favicon: 'Remove favicon' }
+        expect(response).to redirect_to('/admin/appearance?locale=en')
```
bkiahstroud, line 66: I don't _think_ this is a problem, but if somehow the locale was changed in the test env, would this test fail since it specifies `?locale=en`?

app/uploaders/favicon_uploader.rb
```diff
@@ -0,0 +1,23 @@
...
+    %w[png]
```
bkiahstroud, line 21: Why only `png`? Is that standard for favicons?

app/controllers/hyrax/homepage_controller.rb
labradford: This is the controller action for that file, it's how you set the favicon on windows browsers

spec/controllers/sites_controller_spec.rb
labradford: If the locale is changed in the test, then you would need to update the path here. this pattern is used throughout this test

app/uploaders/favicon_uploader.rb
labradford: good catch. they've added an alternative `extension_allowlist`

app/uploaders/favicon_uploader.rb
labradford: it's universal, so it's the best format because it works across all platforms. if you aren't doing this across different tenants, you would upload several files of different types and map them to different platforms.

labradford: we do have the help text that says the favicon needs to be a square png file

bkiahstroud (approved):

Issue #323, opened 2022-11-30
Hanover College collection issue


labradford:
> We have one tenant reporting that works can't be edited, selecting the edit button produces an error. I can reproduce the error from my [ndroark@anderson.edu](mailto:ndroark@anderson.edu) account. Example work here; https://hanover.hykucommons.org/concern/generic_works/f6a9b912-965f-4b21-8933-550e1c7289fa?locale=en (edited)

https://sentry.io/organizations/scientist-inc/issues/3778430425/?project=6707374&query=&referrer=issue-stream&sort=date&statsPeriod=14d

Seems to be an issue with this collection https://hanover.hykucommons.org/dashboard/collections/29005774-7426-40c5-8555-90bdbdddb9de?locale=en

labradford:
zxcv

labradford:
Error after running script ^^

PR #330, opened 2022-12-01
I323 license edit error

labradford:
# Summary
ref #323 
Production issue after upgrade where editing old works that used old licenses would through errors. This MR fixes the issues with editing works with old CC licenses

# Expected Behavior
Editing an old work with CC3.0 licenses should not through errors


app/services/hyrax/qa_select_service_decorator.rb
```diff
@@ -0,0 +1,24 @@
+# OVERRIDE: Hyrax 3.4.2 fixes issues with empty authority and no implicit conversion of string into array
+
+# frozen_string_literal: true
+
+module Hyrax
+  module QaSelectServiceDecorator
+    
+    def active?(id)
+      unless authority.find(id).empty?
```
jeremyf, line 9: This is likely making two requests; consider the following:

```ruby
result = authority.find(id)
return false if result.empty?
result&.fetch('active') # uncertain why we'd need `&` in this case but included for completeness
```

app/services/hyrax/qa_select_service_decorator.rb
```diff
@@ -0,0 +1,24 @@
...
+        authority.find(id)&.fetch('active')
```
jeremyf, line 10: When I'm reading `fetch` I assume that there will be a key error if there's no `active` key in the result.  Should we have a `fetch('active') { false }`?

app/services/hyrax/qa_select_service_decorator.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+    end
+
+    def include_current_value(value, _index, render_options, html_options)
+      unless value.blank? || active?(value)
```
jeremyf, line 15: Unless and or statements always confound me, so this is my bias coming in; I believe the following is logically the same: `if value.present? && !active?(value)`. (again `unless` with `or` always confounds me).

Given the above, my read is:

"Hey, you have a controlled value that is no longer active so we'll add the `force-select` class and add some render options."

another question is how do we handle labels for non-active values?

app/services/hyrax/qa_select_service_decorator.rb
```diff
@@ -0,0 +1,22 @@
+# OVERRIDE: Hyrax 3.4.2 fixes issues with empty authority and no implicit conversion of string into array
+
+# frozen_string_literal: true
+
+module Hyrax
+  module QaSelectServiceDecorator
+    
+    def active?(id)
+      authority.find(id)&.fetch('active') unless authority.find(id).empty?
```
jeremyf, line 9: One concern I have is in regards to multiple `authority.find(id)`; if the find is expensive we're calling it twice.

I recommend the following:

```
result = authority.find(id)
return false if result.empty?
result.fetch('active') { false }
```

The above method reduces duplicate find calls (which could be expensive).  And given that we're calling `fetch` my assumption is the data is a Hash, which could raise an error if the hash does not have the `"active"` key; hence providing the fallback value (which may or may not be the correct answer for a fallback).

app/services/hyrax/qa_select_service_decorator.rb
labradford: This is code from Hyrax and I don't want to mess with it as we need this in Production ASAP

bkiahstroud (approved):

PR #338, opened 2022-12-05
hiding favicon tab until it works

labradford:
# Summary
ref: #149 
Favicon is erroring in staging. This MR hides the favicon tab for users, while allowing us to debug in staging. 

This will allow a deploy to production without removing all of the favicon code.

### DIFF
--- a/app/helpers/hyrax_helper.rb
+++ b/app/helpers/hyrax_helper.rb
@@ -21,6 +21,10 @@ module HyraxHelper
     Site.instance.banner_image? ? Site.instance.banner_image.url : super
   end
 
+  def favicon
+    Site.instance.favicon? ? Site.instance.favicon.url : super
+  end
+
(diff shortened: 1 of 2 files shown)


app/helpers/hyrax_helper.rb
```diff
@@ -21,6 +21,10 @@ def banner_image
     Site.instance.banner_image? ? Site.instance.banner_image.url : super
   end
 
+  def favicon
+    Site.instance.favicon? ? Site.instance.favicon.url : super
+  end
+
```
bkiahstroud, line 27: @labradford should this PR only hide the tab, or is this needed here?

labradford: i want to see if that code fixes the bug we are experiencing

bkiahstroud (approved):

Issue #288, opened 2022-11-09
Collection-level searching not working


DiemBTran:
When viewing a collection page, the central search bar does not return results for the content or metadata of works or sub-collections, only the titles. Optimally, this search would function just as the main search bar for each tenant does, searching the text of PDFs, and all metadata fields, but only within the collection.

DiemBTran:
tested on:
- tested on work and sub-collection metadata
- https://dev.commons-archive.org/collections/05933519-23d0-4876-a0e8-31680e54263e?utf8=%E2%9C%93&cq=miller&sort=&locale=en
- https://dev.commons-archive.org/collections/03390bb4-5e00-4000-b97f-514a40701712?utf8=%E2%9C%93&cq=&sort=&locale=en

Metadata fields that come up in search:
- [x] Description
- [x] Creator
- [x] Keyword
- [x] Date created

Metadata fields that do not come up in search:
- [ ] Subject
- [ ] Language
- [ ] Keyword
- [ ] Date created
- [ ] Resource type
- [ ] text content of PDFs

DiemBTran:
**This passes SoftServ QA:** See attached screen recording.


tested on:
- https://dev.commons-archive.org/collections/03390bb4-5e00-4000-b97f-514a40701712?utf8=%E2%9C%93&cq=&sort=&locale=en
- 
Metadata fields that come up in search:
- [x] Description
- [x] Creator
- [x] Keyword
- [x] Date created
- [x] Subject
- [x] Language
- [x] Keyword
- [x] Resource type
- [x] text content of PDFs

<details><summary> Works' metadata, OCR in works, and Collections' metadata are now searchable from the dashboard collection show page and public collection show page</summary>


https://user-images.githubusercontent.com/29311858/207952225-a66d0ba9-51a3-43fb-a3c3-857df98088d2.mp4
</details>

labradford:
contribute back PR https://github.com/samvera/hyku/pull/1876

PR #348, opened 2022-12-09
I288 collection search

labradford:
# Summary
ref: #288 
On the Collection pages, both admin and public, using the search bar should return a list of documents that match the search term. The search matches should include metadata and full text

# Screenshots / Video
https://share.getcloudapp.com/6quNPKQK

# Expected Behavior
Performing a search on a collection page should return matches to title, metadata and PDF contents


app/search_builders/hyrax/collection_member_search_builder_decorator.rb
```diff
@@ -0,0 +1,39 @@
+# OVERRIDE: Hyrax 3.4.1 adds filesets to search to allow full text search results
+
+# frozen_string_literal: true
+
+module Hyrax
+  module CollectionMemberSearchBuilderDecorator
+    Hyrax::CollectionMemberSearchBuilder.default_processor_chain += [:show_works_or_works_that_contain_files]
+
+    # These methods include the filesets in the search results
+    def show_works_or_works_that_contain_files(solr_parameters)
```
bkiahstroud, line 10: @labradford can you explain the name of this method?

labradford: That's what it was already named in Hyrax here: https://github.com/samvera/hyrax/blob/fd2f525007bedb11a067cec33250a826556f883f/app/search_builders/hyrax/catalog_search_builder.rb#L39

bkiahstroud (approved):

Issue #339, opened 2022-12-06
Enable Google Analytics


DiemBTran:
# Summary

Pals is ready to tinker with Google Analytics.

NEW UA# `UA-251992018-1`

Tested on prod: https://qa.hykucommons.org/. Adding the code `UA-207005751-1` did not immediately enable Google Analytics. Instead, it broke the `Dashboard`,  `Dashboard` > `Collections Report`, and `Dashboard` > `Works Report`. We may need to check if the credentials made it into the code, first. 

# Screenshots or Videos

https://user-images.githubusercontent.com/29311858/205777355-cdfbbf7e-f3c5-4de0-9783-43470e2d477b.mp4

# Testing Instructions

- [ ] As a superadmin, navigate to a tenant's `Dashboard` > `Settings` > `Account` page
- [ ] Enter a Google Analytics code (e.g. `UA-251992018-1`) and save
- [ ] Navigate to  `Dashboard`,  `Dashboard` > `Collections Report`, and `Dashboard` > `Works Report` to verify that those pages are not breaking

DiemBTran:
**This passes SoftServ QA:** See attached screen recording.

tested on:
- https://dev.commons-archive.org/dashboard?locale=en
- https://dev.commons-archive.org/admin/analytics/collection_reports?locale=en
- https://dev.commons-archive.org/admin/analytics/work_reports?locale=en
- https://dev.commons-archive.org/admin/analytics/work_reports/8128c621-23c0-4c6e-9fe3-89d8970e536c?locale=en (not shown in video)
- https://dev.commons-archive.org/works/45ccd215-b514-490b-a253-3374768b5999/stats?locale=en
- NEW UA# UA-251992018-1
- **note:** With the way Google Analytics is currently set up, a Google Analytics code must be added into each tenant for things to function correctly. If a tenant is not given a Google Analytics code, the dashboard will break because there is some Google Analytics functionality on the dashboard page. Things will work differently in the future, so this is a good starting point to begin that discussion so we can estimate the work for those changes 🫡

<details><summary>When a Google Analytics code is added, admin users can access the dashboard, Collections Report, Works Report, Work Report (not shown in video), and Work Show Page Analytics</summary>

https://user-images.githubusercontent.com/29311858/208166354-818b7f8c-c347-4b52-98f0-d4a6b4493da6.mp4</details>

aprilrieger:
[...]

PR #356, opened 2022-12-14
ga-base64-troubleshooting

aprilrieger:
Ref ticket: https://github.com/scientist-softserv/palni-palci/issues/339

- Changing the google analytics after going through the process and setting up a new acct, details located in 1Pass here: https://start.1password.com/open/i?a=LTLZ652TT5H5FHMYMASSH7PIXM&v=huuakin4bu4xanlhktv42qheam&i=fzqgvpva5zgzbazjhcfzn5ei4m&h=scientist.1password.com
- I also found that the server did not print the value of the env variable `GOOGLE_OAUTH_PRIVATE_KEY_VALUE` so we used command substitution instead.
- I am also noting that I have already changed the variables in GitHub.

### DIFF
--- a/bin/web
+++ b/bin/web
@@ -1,6 +1,6 @@
 #!/usr/local/bin/ruby
 if ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] && !ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'].empty?
-  %x{echo #{ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE']} | base64 -d > prod-cred.p12}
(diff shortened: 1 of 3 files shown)


aprilrieger:
base64 path/to/file.p12 | pbcopy
paste that value under <environment staging/demo/production> > secrets > GOOGLE_OAUTH_PRIVATE_KEY_VALUE > Save
Deploy to that <environment>
To verify go to Rancher > Cluster > web pod > execute shell > export > verify the Value of `GOOGLE_OAUTH_PRIVATE_KEY_VALUE` is set and has a value

aprilrieger:
On staging it is working with `UA-251992018-1`

PR #378, opened 2023-01-05
aligning fcrepo values and fedora.yml with hyrax

aprilrieger:
Updating our fedora setup to mirror Hyrax 3.4.2


aprilrieger:
I tested this out on demo and I was still seeing issues with creating tenants with the cname and name of the tenant being created already error. But it really did create the tenant. You log in and you see that there is no solr endpoint because the real error is when the create_account method hits the create admin set step, it errors but we as the user get the above-mentioned error of duplicate account name/cname.

Issue #382, opened 2023-01-09
Push Hyku fixes back into pals


bkiahstroud:
## Summary

While fixing specs in #369, a handful of bugs were found. These are being fixed as part of https://github.com/samvera/hyku/pull/1859. 

Once that PR is merged, the fixes contained within it need to be backported into the pals repo. 

## Acceptance Criteria

- [x] Groups with Roles related bug fixes have been pushed back into pals from Hyku
  - [x] Code changes have been brought into pals (`cherry-pick`ed) 
  - [x] Changes in diff have been reviewed and verified (except specs)
  - [x] Make any additional changes necessary (bringing overrides, configs, etc. inline with updates) 
  - [x] Ensure new decorator overrides merge with existing `class_eval` overrides 
  - [x] Make sure no custom pals code has been lost 
  - [x] Verify translations 
  - [x] Fix failing specs 
  - [ ] ~~Fix rubocop~~ (split to #418) 
  - [ ] ~~Skip labeled tests in CI~~ (split to #417) 

## Testing Instructions 

1. Navigate to an existing tenant 
2. Sign in as an admin 
3. Navigate to Dashboard > Collections > All Collections. Look through each Admin Set until you find one with a mediated deposit Workflow. If none exist, create a new Admin Set and give it a mediated deposit workflow 
4. Navigate to Works. Click the Create New Work button and select any work type 
5. Make the work Public. Fill out the required fields 
6. On the Relationships tab, select the Admin Set with a mediated deposit workflow (see Step 4) 
7. Click Save 
8. Navigate to Dashboard > Review Submissions 
9. Verify your new work shows up in the Under Review tab 
10. Click your work's title 
11. In the Review and Approval dropdown, select Approve. Click Submit 
12. Verify the work moves into the Deposited state 
13. In a private/incognito tab, search for your work's title. Verify you can see the work 
14. Repeat steps 2-13 in a newly created tenant

bkiahstroud:
Got [this error](https://scientist-inc.sentry.io/issues/3623612603/?query=is%3Aunresolved&referrer=issue-stream) when approving a Public `GenericWork` on a new tenant (deploy)

DiemBTran:
[...]

PR #415, opened 2023-02-09
i382 Bring in Hyku fixes for Groups with Roles

bkiahstroud:
Ref https://github.com/scientist-softserv/palni-palci/issues/382 

# Summary

`cherry-pick`ed and resolved merge conflicts from this commit in Hyku: ac8f314e67a3b6edc9965877c9f2911c1c88b4a7


app/models/ability.rb
```diff
@@ -4,7 +4,7 @@ class Ability
   include Hydra::Ability
   include Hyrax::Ability
   include GroupAwareRoleChecker
-  # OVERRIDE: Added custom user ability roles
+  # Add custom ability roles
```
aprilrieger, line 7: I don't see override here, is that on purpose?

bkiahstroud: Yes. Since it's adding behavior, not changing behavior that already exists, Rob and I agreed that it wasn't actually an override (you'll see a few other comments throughout this PR where other "OVERRIDE"s are removed for the same reason)

labradford (approved):

Issue #421, opened 2023-02-28
A single non DC field is included in the OAI feed


crisr15:
# Summary

Using Jio or Adventist as a reference make one field for a non DC term appear in the OAI feed.  

# Acceptance Criteria

- [ ] A new OAI metadata prefix exists to support non DC terms.  (maybe name it hyku for advertising purposes???) add the metadata prefix to the testing instructions for this ticket
- [ ] Select a non DC field that looks difficult and add it to the testing criteria for this ticket.  
- [ ] When I go to see the OAI feed on a record, I can see the term from above that is not a DC term in the feed.  

# Testing instructions

1. select a tenant and use the documentation provided above and this playbook article to access the OAI feed.
2. Select a record from the tenant and look it up in the OAI feed using the metadata prefix oai_hyku
3. Confirm term is showing in the feed

# Notes

Jio and Adventist are utilizing adaptions to the OAI outgoing gem that allows for all metadata fields instead of just Dublin Core Metadata. Rob, Jeremy, or Braydon may have additional insight.

Standard format for oai url in hyku is: https://dc.utk-hyku-staging.notch8.cloud/catalog/oai?verb=ListRecords&metadataPrefix=oai_dc
Playbook article to look at OAI feeds: https://playbook-staging.notch8.com/en/samvera/oai-feeds

Part 1 of Epic #403

bkiahstroud:
PR: https://github.com/scientist-softserv/palni-palci/pull/429

bkiahstroud:
`based_near` ("Location") is probably the most complicated field to display in OAI. However, PALs does not use `based_near`. We need to check this field works when contributing back to Hyku

bkiahstroud:
This is working, but we have a question for Rob in Slack [here](https://assaydepot.slack.com/archives/C03B6A93E31/p1677889840237369)

DiemBTran:
**This passes SoftServ QA:** See attached screenshot. I compared the terms in the `oai_dc feed` to the terms in the `oai_hyku` feed. The `oai_hyku` feed has non-dc feed terms.

tested on:
- `oai_hyku` feed: https://bloomsburg.commons-archive.org/catalog/oai?verb=ListRecords&metadataPrefix=oai_hyku
- `oai_dc` feed: https://bloomsburg.commons-archive.org/catalog/oai?verb=ListRecords&metadataPrefix=oai_dc

<details><summary>The oai_hyku feed has non-dc feed terms.</summary>

[...]

PR #429, opened 2023-03-03
i421 Custom OAI prefix and feed

bkiahstroud:
Ref https://github.com/scientist-softserv/palni-palci/issues/421

# Summary

OAI will now display all public Dublin Core ("DC") and non-DC fields using the new `oai_hyku` metadata prefix. 

# Screenshots / Video

OAI 2 0 Request Results 2023-03-07 at 9 17 11 AM

# Expected Behavior

When I navigate to `/catalog/oai?verb=ListRecords&metadataPrefix=oai_hyku` within a tenant, I expect to be able to see all public metadata properties of all public works. This applies to all custom public properties on all work types.


lib/oai/provider/metadata_format/hyku_dublin_core.rb
```diff
@@ -0,0 +1,60 @@
...
+
+module OAI
+  module Provider
+    module MetadataFormat
+      class HykuDublinCore < OAI::Provider::Metadata::Format
+        def initialize
+          @prefix = 'oai_hyku'
+          @schema = 'http://dublincore.org/schemas/xmls/qdc/dcterms.xsd'
+          @namespace = 'http://purl.org/dc/terms/'
+          @element_namespace = 'hyku'
```
bkiahstroud, line 11: Are these three values appropriate / correct?

lib/oai/provider/metadata_format/hyku_dublin_core.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+
+        def header_specification
+          {
+            'xmlns:oai_dc' => "http://www.openarchives.org/OAI/2.0/oai_dc/",
+            'xmlns:oai_hyku' => "http://www.openarchives.org/OAI/2.0/oai_dc/",
+            'xmlns:dc' => "http://purl.org/dc/elements/1.1/",
+            'xmlns:dcterms' => "http://purl.org/dc/terms/",
+            'xmlns:xsi' => "http://www.w3.org/2001/XMLSchema-instance",
+            'xsi:schemaLocation' => "http://dublincore.org/schemas/xmls/qdc/dcterms.xsd"
```
bkiahstroud, line 52: Are these three values appropriate / correct?

lib/oai/provider/metadata_format/hyku_dublin_core.rb
orangewolf: this is fine. it's not exactly "proper" but its the right thing to do in order to map correctly in other systems.

orangewolf (approved):

Issue #410, opened 2023-01-30
Analytics - File level analytics report


labradford:
# Summary
references https://github.com/scientist-softserv/palni-palci/issues/391

The analytics button on the file level returns a 500 error (example: https://au-archives.hykucommons.org/concern/parent/50c34d7b-5749-40e9-977a-0de2f8ccd58c/file_sets/20792108-3d88-4c8c-ab89-aae824aee6c5)

https://sentry.io/organizations/scientist-inc/issues/3871723870/?project=6707374&query=is%3Aunresolved&referrer=issue-stream

# Acceptance Criteria

- [ ] When clicking the analytics button on the file set show page, the report does not result in a 500 error page

# Screenshots or Video

# Testing Instructions

# Notes
this is not an issue across all filesets, but is an issue with this example.

this issue can be resolved by downloading the fileset and then re-trying the analytics button. This might be a nil check issue.

start in Hyrax in `app/models/statistics.rb` 
Method doesn't seem to exist on the file set level

alishaevn:
it appears that this issue was resolved on https://github.com/samvera/hyrax/issues/5955 and is included in the [hyrax-v4.0.0.rc1 ](https://github.com/samvera/hyrax/releases/tag/hyrax-v4.0.0.rc1) pre release.

DiemBTran:
**This passes SoftServ QA:** See attached screenshot.

tested on:
- work: https://dev.commons-archive.org/concern/images/cba09dd2-9114-4131-bbdd-5d2a061c7ef2?locale=en
- file set: https://dev.commons-archive.org/concern/parent/cba09dd2-9114-4131-bbdd-5d2a061c7ef2/file_sets/31bf949b-1596-4e6d-9dfb-2bc6b807711a
- file level analytics: https://dev.commons-archive.org/files/31bf949b-1596-4e6d-9dfb-2bc6b807711a/stats?locale=en

<details><summary>File level analytics report no longer throws a 500 error</summary>

![File Stat Hyku Commons 2023-05-17 at 1 38 47 PM](https://github.com/scientist-softserv/palni-palci/assets/29311858/41e310db-f8a5-4a9b-980c-cf47cd458f23)
</details>

PR #477, opened 2023-05-16
410 file analytics

alishaevn:
# Story
- ref #410 

We needed to change the method to reference a class that exists, ensuring that the class gets required.

Legato dynamically creates methods based on classes/modules that extend it’s model.  And depending on the load sequence of classes, these dynamic methods can vary. By adding an explicit require, we ensure that those methods are properly defined on the profile.

trying to view the analytics for a file would fail

we can view file analytics

# Screenshots / Video
image

# resources
- https://github.com/tpitale/legato/blob/master/lib/legato/profile_methods.rb#L17-L19
- https://github.com/samvera/hyrax/issues/5955


alishaevn:
@orangewolf @jeremyf and I would like your feedback on the solution here. it incorporates the changes from [hyrax#5958](https://github.com/samvera/hyrax/pull/5958), but the code still failed without the `require` statements because the `Legato::ProfileMethods.instance_methods` weren't finding `hyrax__pageviews` or `hyrax__downloads`.

do you have a better solution?

app/models/file_download_stat_decorator.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax hyrax-v3.5.0 to require Hyrax::Download so the method below doesn't fail
+
+require 'hyrax/download'
```
orangewolf, line 5: I _think_ you want this to require_dependency here https://apidock.com/rails/ActiveSupport/Dependencies/Loadable/require_dependency

orangewolf (changes requested):

alishaevn:
rob and I paired on the refactor. merging.

Issue #409, opened 2023-01-30
Analytics - Works report


labradford:
# Summary
reference issue https://github.com/scientist-softserv/palni-palci/issues/391

In a new, unpublished tenant with no works, the works report is showing view and download numbers.

# Acceptance Criteria

- [ ] In the works analytics report, works will only show the view and download numbers for works in that tenant

# Testing Instructions
Create (or look at) a new tenant with no works.  Confirm that under the analytics -> works report that you can not see any results of works.  

# Notes
in Hyrax, `app/controllers/hyrax/admin/analytics/works_reports_controller` looks like it is querying all of the works and not just works for a tenant. This would need to be fixed in Hyku, as Hyrax does not have tenants.

alishaevn:
this work (along with #408) will be implemented in 3 phases. overall, we've moved analytics to a per tenant system instead of per application. this does mean that deleted works will show in a tenant, for historical purposes. however, by the end of phase 3 if you see a deleted work it is because it was deleted under that tenant's analytics account, not because it's trying to pull data from another tenant. which is currently the case.

## phase one
add analytics settings fields on the tenant level: dashboard >> settings >> account >> (look for fields that begin with "analytics"). these fields need to be filled out for **each** tenant where analytics are desired, with their own account info so that analytics are specific to that tenant. refer to [these instructions](https://github.com/scientist-softserv/palni-palci#analytics-feature) on how to get/create these values.

in order to not lose any tracking time, tenants without the above settings will continue using the account that is currently set app wide as a fallback.

## phase two
once you've had a chance to update the chosen tenants on the demo environment, we will remove the fallback system for the demo environment. any tenants without analytics settings at the tenant level, will no longer have analytic tracking set up. any analytics partials (like the user graph on the dashboard, etc.) will no longer show for them. the tenants with analytics settings will continue as is.

upon your client qa approval, we will also push the phase one code changes to production so that those tenants can be updated.

## phase three
once you've had a chance to update the chosen tenants on the production environment, we will remove the fallback system for the production environment. any tenants without analytics settings at the tenant level, will no longer have analytic tracking set up. any analytics partials (like the user graph on the dashboard, etc.) will no longer show for them. the tenants with analytics settings will continue as is.

alishaevn:
## phase 1 softserv qa
settings were applied to the dev.commons-archive.org tenant on demo.
[...]

Issue #408, opened 2023-01-30
Analytics - Collection Report shows deleted collections


labradford:
# Summary
references issue https://github.com/scientist-softserv/palni-palci/issues/391

When I look at the Collections Report for individual tenants, many (not all) of them share identical lists of "Collection deleted" collections with identical numbers, including brand new tenants that have never had collections created. Is this data that will be gradually replaced with accurate per-tenant data as views start to roll in with the analytics enabled, or is it something worry about? (The works, on the other hand, simply show no files, which is what I would have expected for collections as well.)

# Acceptance Criteria

- [ ] When a tenant has no collections, the collections report will not display any collections.
- [ ] When a tenant has collections, you do not see 'collection deleted' collections.  

# Screenshots or Video
![Screenshot 2023-04-24 at 12 10 13 PM](https://user-images.githubusercontent.com/11359350/234093039-7162f17e-ac44-4d5a-85d6-0cb1593dabd9.png)

# Testing Instructions

- [ ] Go to a tenant that has Deleted collections, and go to the collection report under analytics (screenshot from commons archive Notch8 tenant). 
- [ ] Confirm you can not see Deleted collections.  
- [ ] Find a tenant without collections, Go to collection report under analytics and confirm you can not see any deleted collections.  

# Notes
In Hyrax `app/controllers/hyrax/admin/analytics/collection_reports_controller.rb`
this is a rescue from the collections being nil. May need to put a nil check here

aprilrieger:
https://dev.commons-archive.org/admin/analytics/collection_reports?locale=en

aprilrieger:
@labradford for the acceptance criteria of: When a tenant has no collections, the collections report will not display any collections.
- Find a tenant without collections, Go to collection report under analytics and confirm you can not see any deleted collections.

What should this read: Find a tenant without collections, Go to collection report under analytics and confirm you can not see the --->Top collection partial?

labradford:
Find a tenant without collections, Go to collection report under analytics and confirm you can not see any collections, deleted or otherwise

aprilrieger:
desired behavior when a collection has no collections to show:: show the partial but no collections present. slack convo link: https://assaydepot.slack.com/archives/C03B6A93E31/p1683247704451829

alishaevn:
[...]

PR #479, opened 2023-05-17
408 analytics per tenant

alishaevn:
# Story
- fixes: #408
- fixes: #409 
- related: https://github.com/samvera/hyrax/pull/6063
- related: #482
- related: #487 

- analytics were set through the `.env` or deploy.tmpl files
- one analytics account related to all tenants
- the feature is on by default

- analytics info has to be set per tenant
- the feature is off by default

# Screenshots / Video

<details>
<summary>analytics by default</summary>

Screenshot 2023-05-17 at 1 09 51 PM
Screenshot 2023-05-17 at 1 09 04 PM
</details>

<details>
<summary>analytics enabled</summary>

Screenshot 2023-05-17 at 1 06 17 PM
Screenshot 2023-05-17 at 1 03 57 PM
</details>


app/services/hyrax/analytics/google.rb
```diff
@@ -0,0 +1,214 @@
+# frozen_string_literal: true
+
+require 'oauth2'
+require 'signet/oauth_2/client'
+
+# OVERRIDE: Hyrax hyrax-v3.5.0 to make analytics a bit more dynamic
+# TODO(alishaevn): remove once https://github.com/samvera/hyrax/pull/6063 is available in our current hyrax version
```
alishaevn, line 7: for this reason, rob said to not make a decorator for this file.

config/analytics.yml
```diff
@@ -6,6 +6,7 @@ analytics:
     analytics_id: <%= ENV['GOOGLE_ANALYTICS_ID'] %>
     app_name: <%= ENV['GOOGLE_OAUTH_APP_NAME'] %>
     app_version: <%= ENV['GOOGLE_OAUTH_APP_VERSION'] %>
-    privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH'] %>
+    privkey_value: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] %>
```
aprilrieger, line 9: If you take a look at the worker, the way this works on a deployment we need the path to be setup and not the value.

alishaevn: hmm. I believe I should delete that then since we won't have analytics related env variables. that shouldn't cause a problem right?

aprilrieger: I am not sure, did you have issue navigating around in the dashboard after a dc down -v and a dc build?

I know I had an issue with it off main, and altered the worker script to add the `ENV['HYRAX_ANALYTICS']` to the if clause so the worker honors the flipper:

```yaml
if ENV['HYRAX_ANALYTICS'] && ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] && !ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'].empty?
  %x{echo "$GOOGLE_OAUTH_PRIVATE_KEY_VALUE" | base64 -d > prod-cred.p12}
end
```

alishaevn: I'll test this.

alishaevn:
``` bash
docker compose down -v
docker compose build
docker compose up -d && docker compose exec web sh
  bundle exec rails s -b 0.0.0.0 -p 3000
```

after the above, the app still loaded for me. I'll do a build with no cache overnight to be doubly sure though.

alishaevn: I made the changes in commit [b03fe08](https://github.com/scientist-softserv/palni-palci/pull/479/commits/b03fe08c2477d70b9daaae931ff7871118d75c7d), ran `dc down -v` and `dc build --no-cache`. that failed because of a wip file I had introduced. I removed that file and ran `dc build` again, which passed. starting up the app with `dc up` shows that the web and worker containers have started (in my docker dashboard). however, I see no logs for either in my terminal or in the dashboard. see below. I also only get the dory "http 502" error in my browser. I've confirmed dory is running though.

<details>
<summary>dashboard and browser</summary>

image

image
</details>

I don't understand why this is. the values I removed from the web and worker bin files are no longer used or referenced in this branch so it shouldn't cause a problem.

as info: I typically start up the app in a detached state, then manually start the server myself. (as noted in my above comment). so idk if the `dc up` command is also problematic on `main` for me.

.env
```diff
@@ -47,16 +47,5 @@ HYKU_MULTITENANT=true
...
 
-##START## Enable Google Analytics
-# Uncomment to enable and configure Google Analytics, see README for instructions.
-HYRAX_ANALYTICS=true
-GOOGLE_ANALYTICS_ID=see-1pass
-GOOGLE_OAUTH_APP_NAME=palni-palci-demo
-GOOGLE_OAUTH_APP_VERSION=1.0
-GOOGLE_OAUTH_PRIVATE_KEY_SECRET=notasecret
-GOOGLE_OAUTH_PRIVATE_KEY_PATH=cred.p12
-GOOGLE_OAUTH_CLIENT_EMAIL=palni-palci-demo@palni-palci-demo.iam.gserviceaccount.com
```
alishaevn, line 58: analytics are determined on a per tenant basis, so these are no longer necessary.

app/services/hyrax/analytics/google.rb
```diff
@@ -0,0 +1,217 @@
...
+
+module Hyrax
+  module Analytics
+    module Google
+      extend ActiveSupport::Concern
+      # rubocop:disable Metrics/BlockLength
+      class_methods do
+        # Loads configuration options from config/analytics.yml for Hyrax apps only. You only need PRIVATE_KEY_PATH or
+        # PRIVATE_KEY_VALUE. VALUE takes precedence.
```
alishaevn, line 16: note that this is for hyrax only. and fix spelling errors.

app/services/hyrax/analytics/google.rb
```diff
@@ -0,0 +1,217 @@
...
+        # Expected structure:
+        # `analytics:`
+        # `  google:`
+        # `    app_name: <%= ENV['GOOGLE_OAUTH_APP_NAME']`
+        # `    app_version: <%= ENV['GOOGLE_OAUTH_APP_VERSION']`
+        # `    privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH']`
+        # `    privkey_value: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE']`
+        # `    privkey_secret: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_SECRET']`
+        # `    client_email: <%= ENV['GOOGLE_OAUTH_CLIENT_EMAIL']`
```
alishaevn, line 25: although I deleted the "analytics.yml" file for this app (as can happen with other hyku apps with this implementation), this file belongs to hyrax. so I left this comment.

app/services/hyrax/analytics/google.rb
```diff
@@ -0,0 +1,217 @@
...
+            @config = config
+          end
+
+          # @return [Boolean] are all the required values present?
+          def valid?
+            config_keys = @config.keys
+            # OVERRIDE: Hyrax hyrax-v3.5.0 to require either setting
+            return false unless config_keys.include?('privkey_value') || config_keys.include?('privkey_path')
+
+            REQUIRED_KEYS.all? { |required| config_keys.include?(required) && @config[required].present? }
```
alishaevn, line 64: without `&& @config[required].present?`, this would return true simply because the keys existed on the config. although they were all empty strings.

config/analytics.yml
alishaevn: bringing up the app in a detached state and manually starting the server, works.

alishaevn: april tested this locally and it worked for her so I deployed this branch to the staging env. it worked! no analytics are set by default, as expected.

image

app/services/hyrax/analytics/google.rb
orangewolf: that makes sense, but that means we should change line 62

app/services/hyrax/analytics/google.rb
```diff
@@ -0,0 +1,217 @@
...
+
+          def initialize(config)
+            @config = config
+          end
+
+          # @return [Boolean] are all the required values present?
+          def valid?
+            config_keys = @config.keys
+            # OVERRIDE: Hyrax hyrax-v3.5.0 to require either setting
+            return false unless config_keys.include?('privkey_value') || config_keys.include?('privkey_path')
```
orangewolf, line 62:
```suggestion
            return false unless @config['privkey_value'].present? || @config['privkey_path'].present?
```

app/services/hyrax/analytics/google.rb
orangewolf, line 64:
```suggestion
            REQUIRED_KEYS.all? { |required| @config[required].present? }
```

orangewolf: actually I think the config_keys.include? is reduant. its either present or missing or blank and we don't care which of the later two it is.

bin/web
```diff
@@ -1,6 +1,2 @@
 #!/usr/local/bin/ruby
-if ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] && !ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'].empty?
```
orangewolf, line 2: @alishaevn this needs to be a separate PR. You want to go to production with this still working and THEN fill in the tenants. once that is done, you can remove this as a different PR -> deploy step.

bin/worker
```diff
@@ -1,8 +1,4 @@
 #!/usr/local/bin/ruby
-if ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] && !ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'].empty?
```
orangewolf, line 2: @alishaevn this needs to be a separate PR. You want to go to production with this still working and THEN fill in the tenants. once that is done, you can remove this as a different PR -> deploy step.

config/analytics.yml
```diff
@@ -1,11 +1,12 @@
 #
 # To integrate your app with Google Analytics, uncomment the lines below and add your API key information.
 #
-analytics:
```
orangewolf, line 4: this template should still exists, we can just unset the variables it if we do not want tenants to fall back to a default. I'm not even sure we do want that?

config/browse_everything_providers.yml
```diff
@@ -12,8 +12,8 @@ file_system:
 #   :client_id: YOUR_BOX_CLIENT_ID
 #   :client_secret: YOUR_BOX_CLIENT_SECRET
 # google_drive:
-#   :client_id: YOUR_GOOGLE_API_CLIENT_ID
```
orangewolf, line 15: this is incorrect. Google drive and Google analytics are two different things.

config/locales/simple_form.en.yml
```diff
@@ -13,6 +13,13 @@ en:
           url: Fedora URL should not end with a slash
         name: A single or hyphenated name used for technical aspects of the repository (e.g., "acme" or "acme-library").
         gtm_id: The ID of your Google Tag Manager account
+        analytics_id: The ID of your Analytics account
+        analytics_oauth_app_name: The name of the application
```
orangewolf, line 17:
```suggestion
        analytics_oauth_app_name: The name of the Google application in the Google API console
```

config/locales/simple_form.en.yml
```diff
@@ -13,6 +13,13 @@ en:
...
+        analytics_oauth_app_version: The version of application
```
orangewolf, line 18:
```suggestion
        analytics_oauth_app_version: The version of Google application in the Google API console
```

config/locales/simple_form.en.yml
```diff
@@ -13,6 +13,13 @@ en:
...
+        analytics_oauth_private_key_secret: The secret provided when you created the key
```
orangewolf, line 19: can you review these last 3 with the idea of "by who and where?" added in? there is a lot implied by these as terse as they are.

ops/staging-deploy.tmpl.yaml
```diff
@@ -86,19 +86,19 @@ extraEnvVars: &envVars
     value: http://fcrepo.staging-fcrepo.svc.cluster.local:8080/rest
   - name: GOOGLE_FONTS_KEY
     value: $GOOGLE_FONTS_KEY
-  # - name: GOOGLE_ANALYTICS_ID
-  #   value: $GOOGLE_ANALYTICS_ID  # Set in GitHub Environment Variables Section
-  # - name: GOOGLE_OAUTH_APP_NAME
+  # - name: ANALYTICS_ID
```
orangewolf, line 89: I don't feel like I understand this change?

orangewolf (changes requested):

ops/staging-deploy.tmpl.yaml
alishaevn: @orangewolf the idea was that since this code is going back into hyku, to not limit the person to using google analytics. even though that is what pals is using. so instead of prefacing everything with "google", we leave it more open.

the change in this file specifically is to match what we have in "account_settings.rb" and everywhere else now.

bin/worker
alishaevn: ok, I'll move this to #482.

bin/web
alishaevn: ok, I'll move this to #482.

config/analytics.yml
alishaevn: you mean we don't want the fallback? I agree, and had deleted this file actually. but it fails at [self.load_from_yaml](https://github.com/scientist-softserv/palni-palci/blob/54a7b255a1d71ac55da0e2ba624d0d264fdcf2b0/app/services/hyrax/analytics/google.rb#L33-L34) since the file doesn't exist. I saw that we have [analytics.yml commented out](https://github.com/samvera/hyku/blob/main/config/analytics.yml) in base hyku, and do not have a "google.rb" file/decorator, so I commented the file out too instead of trying to get around the failure.

would you prefer I try to get around the error?

config/locales/simple_form.en.yml
alishaevn: > by who

I pulled these from what was in the readme and left them vague since this work is going back to hyku, which can also use matomo. and potentially other applications in the future? 

> and where

I wasn't sure where these things could be found for matomo. 
<hr>

would you prefer I explicitly state google and matomo for now? and look up where these things can be found for matomo?

config/locales/simple_form.en.yml
alishaevn: ref: https://github.com/scientist-softserv/palni-palci/pull/479#discussion_r1198129112

config/locales/simple_form.en.yml
alishaevn: ref: https://github.com/scientist-softserv/palni-palci/pull/479#discussion_r1198129112

ops/staging-deploy.tmpl.yaml
alishaevn: although they will all get removed too on #482

bin/worker
alishaevn: moved!

bin/web
alishaevn: moved!

config/browse_everything_providers.yml
alishaevn: reverted!

bin/worker
alishaevn: moved this to #487 instead.

orangewolf (approved):

config/locales/simple_form.en.yml
alishaevn: ref: #492

PR #482, opened 2023-05-17
408 disable analytics defaults - staging/demo

alishaevn:
# Story
we need to wait until #479 has been deployed to staging and the client has set up analytics settings on each applicable tenant before deploying this work.

otherwise, the client shouldn't see any change.

refs:
- #408 
- #409
- #479

- default analytics values are still being set in staging/demo

- tenants without analytics settings will no longer have any analytics features


bkiahstroud (approved):

labradford:
Please do not merge until the pipeline is green

PR #484, opened 2023-05-17
adds UV config to disable downloads based on Account Setting.

sephirothkod:
adds custom config.json for the UV
adds the copying of custom jsons to the yarn build package
adds a decorator to passthrough the custom config json based on the `allow_downloads` account setting

### DIFF
--- a/app/helpers/hyrax/iiif_helper_decorator.rb
+++ b/app/helpers/hyrax/iiif_helper_decorator.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+#
+# Overrides to add download restricted specific uv-config
+module Hyrax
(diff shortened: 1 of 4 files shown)


app/helpers/hyrax/iiif_helper_decorator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+#
+# Overrides to add download restricted specific uv-config
```
alishaevn, line 3:
```suggestion
# OVERRIDE: Hyrax hyrax-v3.5.0 to add download restricted specific uv-config
```

alishaevn (changes requested):

alishaevn (approved):

PR #488, opened 2023-05-18
Restrict downloads

sephirothkod:
wraps download links in an `if` statement that checks if `allow_downloads` is disabled


labradford (changes requested):
Can you change the target branch from `main` to `reshare-feature`

labradford (approved):

PR #494, opened 2023-05-22
408-for-the-last-time

alishaevn:
# Story

Refs
- #408 
- #409 

- without setting tenant level account settings, the dashboard would load fine
- after adding and saving the account settings, the dashboard would err
  - this is because the user graph needs the correct [profile](https://github.com/scientist-softserv/palni-palci/blob/main/app/services/hyrax/analytics/google.rb#L112-L119). `Hyrax::Analytics.config.valid?` would return false however. while `Hyrax.config.analytics` would be true. (this may be a side effect of trying to have a fallback only, since `HYRAX_ANALYTICS=true` is set in the ENV variables)
  - either way, the correct settings would be set on the `Account.settings` object. 

- usw `self` to make sure that everything gets set properly when reloading the analytics properties
- a user is able to see analytics related partials after tenant level analytics have been applied

# Screenshots / Video

image

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -202,13 +202,13 @@ module AccountSettings
              self.google_oauth_client_email.present?
 
         config = Hyrax::Analytics::Config.load_from_yaml
-        self.google_analytics_id ||= config.analytics_id
-        self.google_oauth_app_name ||= config.app_name
-        self.google_oauth_app_version ||= config.app_version
-        self.google_oauth_private_key_value ||= config.privkey_value
-        self.google_oauth_private_key_path ||= config.privkey_path
-        self.google_oauth_private_key_secret ||= config.privkey_secret
-        self.google_oauth_client_email ||= config.client_email
+        self.google_analytics_id = self.google_analytics_id.presence || config.analytics_id
+        self.google_oauth_app_name = self.google_oauth_app_name.presence || config.app_name
+        self.google_oauth_app_version = self.google_oauth_app_version.presence || config.app_version
+        self.google_oauth_private_key_value = self.google_oauth_private_key_value.presence || config.privkey_value
+        self.google_oauth_private_key_path = self.google_oauth_private_key_path.presence || config.privkey_path
+        self.google_oauth_private_key_secret = self.google_oauth_private_key_secret.presence || config.privkey_secret
+        self.google_oauth_client_email = self.google_oauth_client_email.presence || config.client_email
       end
 
       # require the analytics to be set per tenant
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -192,6 +192,7 @@ module AccountSettings
     end
 
     def reload_analytics
+      # rubocop:disable Style/RedundantSelf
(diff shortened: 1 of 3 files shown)


sephirothkod (dismissed):

config/analytics.yml
```diff
@@ -7,6 +7,6 @@ analytics:
     app_name: <%= ENV['GOOGLE_OAUTH_APP_NAME'] %>
     app_version: <%= ENV['GOOGLE_OAUTH_APP_VERSION'] %>
     privkey_value: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_VALUE'] %>
-    # OR privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH'] %>
+    privkey_path: <%= ENV['GOOGLE_OAUTH_PRIVATE_KEY_PATH'] %>
```
alishaevn, line 10: since we are leaving the possibility of setting `self.google_oauth_private_key_value` in `#reload_analytics`, we need the app to know what `config.privkey_path` is.

app/models/concerns/account_settings.rb
```diff
@@ -192,33 +192,36 @@ def reload_library_config
...
-        google_oauth_private_key_secret ||= config.privkey_secret
-        google_oauth_private_key_value ||= config.privkey_value
-        google_oauth_client_email ||= config.client_email
+        self.google_analytics_id = self.google_analytics_id.presence || config.analytics_id
+        self.google_oauth_app_name = self.google_oauth_app_name.presence || config.app_name
+        self.google_oauth_app_version = self.google_oauth_app_version.presence || config.app_version
+        self.google_oauth_private_key_value = self.google_oauth_private_key_value.presence || config.privkey_value
+        self.google_oauth_private_key_path = self.google_oauth_private_key_path.presence || config.privkey_path
+        self.google_oauth_private_key_secret = self.google_oauth_private_key_secret.presence || config.privkey_secret
+        self.google_oauth_client_email = self.google_oauth_client_email.presence || config.client_email
```
alishaevn, line 211: guard against the account setting value being an empty string.

if an account once had analytics settings that have been removed, the value for each would be an empty string. when using the or/equal operator with an empty string and non empty string, the empty string is returned. we cannot however check for the presence of a variable while also using the or/equal operator.

app/models/concerns/account_settings.rb
```diff
@@ -192,33 +192,36 @@ def reload_library_config
...
     def reload_analytics
+      # rubocop:disable Style/RedundantSelf
       # fall back to the default values if they aren't set in the tenant
-      unless google_analytics_id.present? &&
-             google_oauth_app_name.present? &&
-             google_oauth_app_version.present? &&
-             google_oauth_private_key_secret.present? &&
-             google_oauth_client_email.present? &&
-             (google_oauth_private_key_value.present? || google_oauth_private_key_path.present?)
+      unless self.google_analytics_id.present? &&
```
alishaevn, line 197: needed to preface everything with self because of the below

``` bash
google_analytics_id => ''
Account.settings['google_analytics_id'] => 'UA-12345'
```

sephirothkod (approved):

PR #499, opened 2023-05-23
iiif decorator update

sephirothkod:
changed decorator to the iiif_print one.


app/helpers/iiif_print/iiif_helper_decorator.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# OVERRIDE Hyrax v2.9.6 add #uv_search_param
```
labradford, line 3: we are on a newer version of Hyrax than this

labradford: Can you update the comment with override IIIF_Print gem

labradford (approved):

PR #535, opened 2023-06-05
695-update-resource-allocations

aprilrieger:
Refs https://github.com/scientist-softserv/dev-ops/issues/695

Please read the initial issue linked above

### DIFF
--- a/ops/production-deploy.tmpl.yaml
+++ b/ops/production-deploy.tmpl.yaml
@@ -5,11 +5,11 @@ imagePullSecrets:
 
 resources:
   limits:
-    memory: "12GB"
-    cpu: "0.5"
+    memory: "4GB"
+    cpu: "0.6"
   requests:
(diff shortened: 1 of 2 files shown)


ops/production-deploy.tmpl.yaml
```diff
@@ -5,11 +5,11 @@ imagePullSecrets:
 
 resources:
   limits:
-    memory: "4Gi"
```
orangewolf, line 8: this should be 2/4GB like other web processes. I can understand matching UTK, but we're not importing massively here so I think the worker can be much smaller as well

orangewolf (changes requested):
I think pals prod can be a lot smaller

aprilrieger:
@orangewolf I updated the values, let me know if these work.

orangewolf (approved):
looks great!

PR #551, opened 2023-06-07
Allow selenium specs to run on m1

summer-cook:
# Story
Selenium specs currently do not run locally on M1 machines, and we are not able to see anything in the browser. This enables a selenium grid ui at `localhost:4444` and allows the Selenium driver to connect there through an http request, because it was failing to do so before due to WebMock disabling all http requests. 

- Users with an m1 chip could not run any specs that used selenium (mostly in feature specs)

- Users with an m1 chip can run all specs
- Users can see the selenium grid ui at `localhost:4444`
- Users can actually see feature specs running in the browser using a vnc at `vnc://localhost:6900`


docker-compose.yml
summer-cook, line 1: I don't know if this should be an actual change, or if this should maybe be in an override file specifically for m1 users?

This needs to be tested on a non m1 machine to ensure it doesn't break anything

README.md
```diff
@@ -67,6 +67,15 @@ The full spec suite can be run in docker locally. There are several ways to do t
 docker-compose exec web rake
 ```
 
+#### Docker with an M1 chip
+- You may run into issues with Selenium with an M1 chip. If so, use the override file provided called `docker-compose-override.yml`
```
summer-cook, line 71: these will potentially need to be updated, depending on if we want an override file or just to change it normally

docker-compose.yml
aprilrieger: tested on intel chip mac and the tests are working. I followed the directions in this pr and was able to get a feature spec to run in the browser.
Screenshot 2023-06-08 at 9 23 02 AM

aprilrieger (approved):

ShanaLMoore:
this is so great to see. Thank you ladies!!

ShanaLMoore:
For context, Summer closed this because we discovered that this project didn't have the updates seen in hyku proper. She applied the changes seen in the following commit and was able to get it working:

https://github.com/scientist-softserv/britishlibrary/commit/533dd5cbff679f2d340c5db4b0fff2190f20ce5c

summer-cook:
> For context, Summer closed this because we discovered that this project didn't have the updates seen in hyku proper. We discovered this because LaRita (who was having the same issue) was able to run feature specs just fine in UTK. She applied the changes seen in the following commit and was able to get it working:
> 
> [scientist-softserv/britishlibrary@533dd5c](https://github.com/scientist-softserv/britishlibrary/commit/533dd5cbff679f2d340c5db4b0fff2190f20ce5c)'
> 
> ref: [slack convo](https://assaydepot.slack.com/archives/C0313NK5NMA/p1686319538669209)

Reopening now because the updates from hyku proper actually cause the pipeline to fail in ci even though the same tests pass locally. 

I am going to try to bring the changes in this branch more in line with hyku proper. I think some of the things that are in the rails helper may actually be needed.

Issue #516, opened 2023-06-02
On Commons-Archive.org sometimes the url does not load when ?locale=en is removed


crisr15:
# Story

Sometimes when you navigate to the url without the ?locale=en, you get a rails error saying the page is unavailable.  We say this previously, but it does not seem like it was completely resolved: https://assaydepot.slack.com/archives/C0311DNF3MG/p1684529517728409?thread_ts=1684528944.268469&cid=C0311DNF3MG

Try to replicate the issue by removing the ?locale=en from the url, but this error seems to be intermittent.  

# Acceptance Criteria

- [ ] We have an idea of what is causing this issue
- [ ] We have resolved it permanently

# Screenshots / Video

<details>
<summary></summary>
<img width="597" alt="Screenshot 2023-06-01 at 5 04 16 PM" src="https://github.com/scientist-softserv/palni-palci/assets/11359350/dcaff78e-7574-4170-8749-7f19ad03deaf">

</details>

# Testing Instructions and Sample Files

-

# Notes

labradford:
https://scientist-inc.sentry.io/issues/4245781432/?project=6707374&query=is%3Aunresolved&referrer=issue-stream&stream_index=6

ndroark:
In staging I still get error messages clicking on works from the search page. The URL is missing the translation tag. When I manually add the tag to the URL the work loads.

labradford:
I'm running into this error only when logged in. Will investigate further

labradford:
I have found that there were a few users that didn't have the preferred_locale set. On the staging site, these were the users that we created via the rails console. I ran a script to update the users without a preferred_locale to now have "en" set.

We need to keep an eye on this to see if some sort of code change caused the preferred_locale to get removed, or if there was a code change the just exposed this. 

Any new users that we accept user invites via rails console should also have the preferred_locale set during this process.

PR #570, opened 2023-06-12
I516 missing locale

labradford:
# Story

Refs #516 

Pages fail to render because the locale is a blank string

This PR adds a default locale in the head tag to see if this fixes the error with locale being blank
Pages do not load
Pages will load
# Screenshots / Video

<details>
<summary></summary>

</details>

# Notes

### DIFF
--- a/app/views/layouts/hyrax.html.erb
+++ b/app/views/layouts/hyrax.html.erb
@@ -1,6 +1,6 @@
 <!-- add body classes to make styling easier -->
 <!DOCTYPE html>
-<html lang="<%= I18n.locale.to_s %>" prefix="og:http://ogp.me/ns#">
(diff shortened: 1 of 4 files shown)


summer-cook (approved):

app/views/layouts/hyrax.html.erb
```diff
@@ -1,6 +1,6 @@
 <!-- add body classes to make styling easier -->
 <!DOCTYPE html>
-<html lang="<%= I18n.locale.to_s %>" prefix="og:http://ogp.me/ns#">
+<html lang="<%= I18n.locale.to_s || :en %>" prefix="og:http://ogp.me/ns#">
```
alishaevn, line 3: if this does work, is there a way to set a default locale, instead of having a fallback?

labradford: the default locale is set in an initializer. It's not being recognized on staging

Issue #569, opened 2023-06-12
update production analytics


alishaevn:
# Story
we're moving the "phase 3" steps from the referenced pr's below into its own ticket. per today's [client meeting](https://docs.google.com/document/d/1ASOWLZhFBeMUw1KTWxAmEdotbvM471RgQfe2BhkQJ_Q/edit), pals will hold off on having the tenants set up their own analytics accounts until the google analytics 4 switch is ready to happen. that way both things can happen at once.

## phase three
once you've had a chance to update the chosen tenants on the production environment, we will remove the fallback system for the production environment. any tenants that didn't have their account settings saved with the defaults or new values, will no longer have analytic tracking set up. any analytics partials (like the user graph on the dashboard, etc.) will no longer show for them. the tenants with analytics settings will continue as is.

ref:
- #408 
- #409 

# Acceptance Criteria
- [ ] In the works analytics report, works will only show the view and download numbers for works in that tenant
- [ ] In the collections analytics report, collections will only show the view and download numbers for works in that tenant

# Testing Instructions and Sample Files
- [ ] save/update the appropriate analytics settings to each applicable tenant

#### for production tenants that had settings saved/updated
- [ ] confirm that the tenant still shows analytics features on the dashboard (e.g., user graph)
- [ ] confirm that the works report still shows works, if there are any
 
#### for production tenants that did not have settings saved/updated
- [ ] confirm that the tenant shows **no** analytics features on the dashboard (e.g., no user graph)
- [ ] confirm that the works report page mentions that analytics must be turned on

# Notes

PR #579, opened 2023-06-22
hide-per-tenant-analytics-on-prod

alishaevn:
use the default analytics as a default on prod only.

the per tenant analytics work was already merged to main and the defaults were hidden on staging. however, pals wants to hold off on deploying that code to production until the ga4 switch has happened so both major changes can happen at the same time.

putting the analytics defaults in a conditional ensures that we can still deploy to production from main.

related to:
- https://github.com/scientist-softserv/palni-palci/issues/569
- #487

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -196,11 +196,11 @@ module AccountSettings
       if Rails.env.production?
         # fall back to the default values if they aren't set in the tenant
         unless self.google_analytics_id.present? &&
-          self.google_oauth_app_name.present? &&
-          self.google_oauth_app_version.present? &&
-          (self.google_oauth_private_key_value.present? || self.google_oauth_private_key_path.present?) &&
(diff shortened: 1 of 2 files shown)


app/models/concerns/account_settings.rb
```diff
@@ -193,6 +193,26 @@ def reload_library_config
 
     def reload_analytics
       # rubocop:disable Style/RedundantSelf
+      if Rails.env.production?
+        # fall back to the default values if they aren't set in the tenant
+        unless self.google_analytics_id.present? &&
+               self.google_oauth_app_name.present? &&
+               self.google_oauth_app_version.present? &&
+               (self.google_oauth_private_key_value.present? || self.google_oauth_private_key_path.present?) &&
```
aprilrieger, line 201: Perfect!

aprilrieger (approved):
Looks greta Alisha, thanks for spear-heading this work!

PR #580, opened 2023-06-23
add-analytics-yml-fallback

alishaevn:
default the env variables in analytics.yml to be empty strings so they do not fail when the env variable is not present

ref:
- #579 
- [slack thread](https://assaydepot.slack.com/archives/C0311DNF3MG/p1687473554171879) outlining the issue that this pr solves

### DIFF
--- a/config/analytics.yml
+++ b/config/analytics.yml
@@ -3,10 +3,10 @@
 #
 analytics:
   google:
-    analytics_id: <%= ENV['GOOGLE_ANALYTICS_ID'] %>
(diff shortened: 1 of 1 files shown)


alishaevn:
I deployed this branch to `staging` to confirm that I can see the home page, a tenant, and the admin dashboard.

I then deployed this branch to `demo` and everything is also working fine there.

I will wait until this afternoon before deploying to `prod`. probably shortly before the reindex that’s scheduled for 5pm est.

aprilrieger (approved):

Issue #524, opened 2023-06-05
Create advanced search page and function with AND, OR, excluded and Exact Text modifiers


crisr15:
# Story

Create advanced search page and function with AND, OR, excluded and Exact Text modifiers.  Pals would like to use the blacklight advanced search gem and have a similar experience to Adventist Digital Library (https://adl.b2.adventistdigitallibrary.org/advanced) .  Please add and map all the metadata fields, but they do not want any of the 'additional attributes'.  [Metadata for hyku commons](https://docs.google.com/spreadsheets/d/18PM2cD9P9ep5qEEYcVZzK1vzfOzli4DBtVppDc8lpDA/edit#gid=0).    

# Acceptance Criteria

- [ ] Search shows an advanced button so users can get to advanced search
- [ ] Advanced search contains all the available metadata fields.  
- [ ] Advanced search is functioning and the modifiers work.   
- [ ] The advanced search button is included in the toolbar in all 3 themes

## fields
- [ ] Title
- [ ] Creator
- [ ] Keyword
- [ ] Rights Statement or Rights
- [ ] Resource Type or Type
- [ ] License
- [ ] Alternative Title
- [ ] Contributor
- [ ] Description
- [ ] Abstract
- [ ] Access Rights or Additional Rights Information
- [ ] Rights notes
- [ ] Publisher
- [ ] Date Created or Date
- [ ] Subject
- [ ] Language
- [ ] Identifier
- [ ] Related URL
- [ ] Source
- [ ] Bibliographic Citation
- [ ] Extent
- [ ] Audience
- [ ] Education Level
- [ ] Learning Resource Type
- [ ] Discipline
- [ ] Rights Holder
- [ ] Size
- [ ] Table Of Contents
- [ ] Accessibility Feature
- [ ] Accessibility Hazard
- [ ] Accessibility Summary
- [ ] Degree
- [ ] Level
- [ ] Grantor
- [ ] Format
- [ ] Advisor
- [ ] Committee member
- [ ] Department

# Screenshots / Video

<details>
<summary></summary>
https://adl.b2.adventistdigitallibrary.org/advanced
</details>

# Testing Instructions and Sample Files

-

# Notes

Are there any fields you want people to be able to search separately? 
Advanced search gem may be going away in the future to be replaces by something else.  When hyku gets to a later version of blacklight we will need to adjust/upgrade this feature.

ndroark:
[...]

PR #581, opened 2023-06-26
Advanced search rework

summer-cook:
# Story
Advanced search needs to have ALL fields, not just the default search fields

# related
- #524 

- some fields were missing

- all fields listed on the ticket #524 will be shown

# Screenshots / Video

<details>
<summary>Updated Form</summary>

screencapture-pals-hyku-test-advanced-2023-06-26-18_06_55

</details>

# Notes
Several of these fields are actually searching the same thing, they are just labeled differently on the form depending on the worktype. I am not able to add duplicates of the `search_fields` so that they will each get their own row/field on the advanced search form, so in these cases I have to combine them into one field with both of their labels. 

- Rights
   - this is just labeled differently for etds - its actually the exact same thing as rights statement, so i combined them
- Type
  - this is also the same field as resource type, just labeled differently for OER & ETD
- Date
  - same field as date created
- Additional Rights Information
  - as far as i can tell this is the same thing as access rights for all work types (it maps to access rights), and a property of access rights doesn't actually map anywhere so its weird that it is showing up on a form? I included it in the label though just in case

Skipping
- administrative notes as it is admin only


app/controllers/catalog_controller.rb
```diff
@@ -181,195 +181,74 @@ def self.modified_field
...
+        field.label = search_field[:label]
+        field.solr_parameters = { "spellcheck.dictionary": search_field[:name] }
+        solr_name = "#{search_field[:name]}_tesim"
+        field.solr_local_parameters = {
+          qf: solr_name,
+          pf: solr_name
+        }
+      end
     end
+    # If there is something additional about a search field that needs to be customized i.e. whether to include in advanced search, or a different solr name, add it below
```
aprilrieger, line 237: Great note, ty!

aprilrieger (approved):

Issue #530, opened 2023-06-05
Work Show/View Page - Restore functionality of geography field and display field on work page


crisr15:
# Story

The Geography field (`based_near`) has been hidden from users as it was not working correctly.  Fix the field and display to users.  

# Acceptance Criteria

- [ ] a user can use the field in the form to search for based near (location) controlled vocabularies
- [ ] works can be saved with a based near (location) field filled in
- [ ] These changes may require of reindex of prod to take effect
- [ ] works that have an existing based near location filled in do not throw an error upon edit after reindexing all works
- [ ] Help text for based near is updated to be more descriptive and actually help the user set up this field if they haven't already 

# Screenshots / Video

<details>
<summary>error that is thrown on save for anything with a based near</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/73361970/eb02f109-2074-40be-878c-ac350fe16f66)

</details>

<details>
<summary>The first place in the stack where the error is thrown:</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/73361970/f64cb223-ed25-407d-90c3-d0fa83847833)

</details>

<details>
<summary> <s>This is the error that is thrown while trying to search for a Based Near (aka location)</s> this is solved by correctly entering your geonames username & configuring your geonames account.</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/73361970/d0eda40d-94ac-4688-b075-4b72972b350c)

</details>

<details>
<summary> <s>In the edit field, it just says loading failed, and has errors in the console</s> this is solved by correctly entering your geonames username & configuring your geonames account.</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/73361970/a662609b-b8aa-495a-ac16-4d4131c0fc08)

</details>

# Testing Instructions and Sample Files
In order for this field to work, you will need to configure the geonames in each tenant. Follow the following steps to do this: 

### Steps to Enable Based Near Controlled Vocabulary in Hyku/Hyrax app  
1. Create a free account on [geonames.org](http://geonames.org/)
2. Go to your account settings at https://www.geonames.org/manageaccount and select the option at the bottom to enable free services. 
3. Go into your hyrax/hyku account settings in the dashboard
4. Enter your geonames username in the hyrax account settings
5. You should now be able to see the list of locations in the based near field inside the forms

Then, to test
1. create a work with the location field filled in, make sure it saves.
2. edit that work and make sure it saves

# Notes
_TODO_
[...]

PR #591, opened 2023-06-28
530 - Bring Back Based Near

summer-cook:
# Story
Fixes/Re-Enables the `based_near` field so that an error is not thrown upon trying to create a work with based near location set

TODO: 
- update the label for this field to location?

# Related
- #530

Based near was turned off

- users can select based near geography location for their works
- users can select multiple locations

# Screenshots / Video
Video
https://share.getcloudapp.com/P8uXkxmg

<details>
<summary>Form field with finished help text (since it was not finished in the video)</summary>
image

</details>

# Notes/Testing Instructions
In order for this field to work, you will need to configure the geonames in each tenant. Follow the following steps to do this: 

### Steps to Enable Based Near Controlled Vocabulary in Hyku/Hyrax app  
1. Create a free account on [geonames.org](http://geonames.org/)
2. Go to your account settings at https://www.geonames.org/manageaccount and select the option at the bottom to enable free services. 
3. Go into your hyrax/hyku account settings in the dashboard
4. Enter your geonames username in the hyrax account settings
5. You should now be able to see the list of locations in the based near field inside the forms

Then, to test
1. create a work with the location field filled in, make sure it saves.
2. edit that work and make sure it saves


lib/hyrax/controlled_vocabularies/location_decorator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module LocationDecorator
+    # the Location class throws an error that split is undefined since it is an enumerator, but this error is coming from deeply nested dependencies.
+    # This causes 500 errors for works with their Location field filled in.
+    # defining split just for the purpose of getting past this error seemed to work and allow locations to save correctly with no adverse effects.
+    # See https://github.com/scientist-softserv/palni-palci/issues/524 for more details
+    def split(*)
```
summer-cook, line 9: This didn't seem to have any adverse effects since it is only applying to this one specific class that is only used for this controlled vocab, but please let me know if there is a better way to do this. I searched in Hyrax but there didn't seem to be any differences for any of the files used for this controlled vocab between the newest version and the version PALS is on, so this error is still very mysterious to me. 

The place where split was being called was so deeply nested it was hard to find. But, it seems like just defining it still allows the field to save with the correct location names.

config/application.rb
```diff
@@ -36,9 +36,15 @@ class Application < Rails::Application
     end
 
     config.to_prepare do
-      # Allows us to use decorator files in the app directory
-      Dir.glob(File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb")).sort.each do |c|
-        Rails.configuration.cache_classes ? require(c) : load(c)
+      decorator_paths = [
+        File.join(File.dirname(__FILE__), "../app/**/*_decorator*.rb"),
+        File.join(File.dirname(__FILE__), "../lib/**/*_decorator*.rb")
+      ]
```
alishaevn, line 42: 👍🏾 I like it!

lib/hyrax/controlled_vocabularies/location_decorator.rb
alishaevn: how does #524 related to this method/file?

summer-cook:
@alishaevn good catch i meant to put #530! not 524

lib/hyrax/controlled_vocabularies/location_decorator.rb
summer-cook: @alishaevn good catch i meant to put https://github.com/scientist-softserv/palni-palci/issues/530! not 524

lib/hyrax/controlled_vocabularies/location_decorator.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Hyrax
+  module LocationDecorator
```
bkiahstroud, line 4: @summer-cook based on the file path and original class name, shouldn't this decorator be: 

```ruby
Hyrax::ControlledVocabularies::LocationDecorator
# instead of just 
Hyrax::LocationDecorator
```

summer-cook: so it would be like 

```
module Hyrax
    module ControlledVocabularies
        module LocationDecorator
```

?

i'm going to try it and let you know if it works

summer-cook: @bkiahstroud ok so both ways work somehow. It seems like the second way is a bit more clear, so will push that up

alishaevn (approved):

labradford:
In the past a branch name with the `/` in it would break the deploy/pipeline, but I don't know if that was a gitlab issue. Just something we should keep an eye on

Issue #599, opened 2023-07-05
Reshare show page theme


summer-cook:
# Story
Reshare show page theme wireframe: https://www.figma.com/file/K4o9KLOKXlGwVDi7kVVeB1/Untitled?type=design&node-id=0-1&mode=design&t=08zh0rJz0t2gOI3M-0

# Acceptance Criteria
- [ ] There is a ReShare show page theme that can be selected from the Appearance>Themes settings
- [ ] ReShare show page theme only has UV and Metadata
- [ ] ReShare show page theme will have links back to ReShare for additional search

# Testing Instructions
- [ ] To test on staging, log into https://reshare.palni-palci-staging.notch8.cloud/ as admin user
- [ ] Navigate to the `Dashboard > Settings > Appearance` page here: https://reshare.palni-palci-staging.notch8.cloud/admin/appearance?locale=en
- [ ] Click the `Themes` tab
- [ ] Scroll down to `Show Page Theme`. You will see a dropdown menu and a wireframe beside it. This wireframe changes depending on which option is selected
- [ ] If it's not already selected, select `ReShare Show Page` on the dropdown menu. Click the `Save changes` button
- [ ] In another window, visit this work show page: https://reshare.palni-palci-staging.notch8.cloud/concern/cdls/7cee6c8f-0fd0-4f0c-88ed-8441fc4d16a7?locale=en
- [ ] Compare this work show page to the `ReShare Show Page` wireframe
- [ ] Verify that the work show page is accurate to the wireframe
  - **Note:** For web accessibility reasons, we changed the footer link font for `Samvera` to be darker. There needs to be sufficient contrast between the word `Samvera` and its background color to meet accessibility standards.
<details><summary>Click to expand and see a larger version of the wireframe and what you're comparing it to. </summary>

![reshare_show-df22fb6abca1d97baf325597860a08fde849f67479992a86a95d0ec40c7cbf3a](https://github.com/scientist-softserv/palni-palci/assets/29311858/61b94fb5-2c7b-4bd3-8a68-7bc3d20ed8d0)
![Cdl Description of the Camera Lucida ID 7cee6c8f-0fd0-4f0c-88ed-8441fc4d16a7 Hyku Commons 2023-07-25 at 6 58 28 PM](https://github.com/scientist-softserv/palni-palci/assets/29311858/d44c12c0-8639-4602-8e9c-055b7c73da91)

</details>

# Related
EPIC: #462 

# Files

[...]

PR #604, opened 2023-07-06
I599 show page theme

summer-cook:
# Story
ReShare Show page theme

# Related
- #599 

- There was no unique theme for ReShare

- [ ] There is a unique theme for ReShare's work-show page
- [ ] There are no links to elsewhere in Hyku for non admins

# Screenshots / Video

<details>
<summary>show page</summary>
image

</details>

# Notes
- we will need to go back and add the link urls for reshare's site for the 2 buttons on this page


app/views/themes/reshare_show/_controls.html.erb
labradford, line 1: I think I would leave a code comment in this file "this file is intentionally left blank" just so someone doesn't come along and delete it by accident

app/views/themes/reshare_show/_user_util_links.html.erb
```diff
@@ -0,0 +1,24 @@
+<ul id="user_utility_links" class="nav navbar-nav navbar-right">
+  <%= render 'shared/locale_picker' if available_translations.size > 1 %>
+  <% if user_signed_in? %>
```
labradford, line 3: we only want admins to see this. other users will be signed in, but we don't want them to see anything except for the viewer + metadata.

app/views/themes/reshare_show/hyrax/base/_attribute_rows.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= presenter.attribute_to_html(:admin_note, html_dl: true) if presenter.editor? %>
```
labradford, line 1: do you need this file? Is there custom metadata for ReShare? Or does this have the same items as the base attribute_rows?

app/views/themes/reshare_show/hyrax/base/_metadata.html.erb
```diff
@@ -0,0 +1,5 @@
+<dl class="work-show <%= dom_class(presenter) %> mb-0" <%= presenter.microdata_type_to_html %>>
+  <%= render 'attribute_rows', presenter: presenter %>
```
labradford, line 2: If you can use the base attribute rows, you can change this to add the path to it. `render '../../hyrax/base/attribute_rows'` or something like that

app/views/themes/reshare_show/hyrax/base/_metadata.html.erb
```diff
@@ -0,0 +1,5 @@
...
+  <%= presenter.attribute_to_html(:embargo_release_date, render_as: :date, html_dl: true) %>
```
labradford, line 3: I'm not sure embargo or leases will apply to reshare. we can probably remove these.

labradford (changes requested):
Some questions/comments for you

app/views/themes/reshare_show/hyrax/base/_attribute_rows.html.erb
summer-cook: It has the same items as base attribute rows, but they are not rendering as faceted as in the base attribute rows since users should not be able to go to any hyku search pages

app/views/themes/reshare_show/hyrax/base/_metadata.html.erb
summer-cook: unfortunately we can't use the base/attribute_rows because of the render as:faceted thing

app/views/themes/reshare_show/_user_util_links.html.erb
summer-cook: Ok sounds good, i just figured the `user_signed_in?` would not apply to them unless they were signed in with the devise (which is different from sso i thought?). I can change it to only admin users though

labradford: you will be able to sign in both ways, the SSO is still through devise.

app/views/themes/reshare_show/_controls.html.erb
summer-cook: done!

labradford (approved):

Issue #600, opened 2023-07-05
Universal viewer theming


summer-cook:
# Summary
Make edits to the UV for reshare show page theme

# Acceptance Criteria
- [ ] There is no download link for the work in the UV
- [ ] UV can show 2 page view

# Testing Instructions
- [ ] To test on staging, log into https://reshare.palni-palci-staging.notch8.cloud as an Admin user
- [ ] Visit this work show page: https://reshare.palni-palci-staging.notch8.cloud/concern/cdls/7cee6c8f-0fd0-4f0c-88ed-8441fc4d16a7?locale=en
- [ ] Verify the Universal Viewer (UV) has its download link removed
<details><summary>Click to expand these screenshots. The first shows a default UV from another app which still has the download link. The second screenshot shows that the ReShare UV has its download link removed</summary>

![Generic Work 2023-07-09 Revert to main branch of IiifPrint #454 ID 94f5d7e5-90d1-44e4-925c-9fc075c6e934 Hyku 2023-07-25 at 6 42 52 PM](https://github.com/scientist-softserv/palni-palci/assets/29311858/c09484fa-e228-4708-93e5-79ad9164d424)
![Cdl Description of the Camera Lucida ID 7cee6c8f-0fd0-4f0c-88ed-8441fc4d16a7 Hyku Commons 2023-07-25 at 6 42 44 PM](https://github.com/scientist-softserv/palni-palci/assets/29311858/dd45c932-709e-4692-9411-61edb7e27b4b)
</details>

- [ ] The first page of the PDF is the front cover and the last page of the PDF is the back cover of the book. Flip through the pages on the UV by clicking the right arrow. Verify that the UV shows 2 pages at a time for every other set of pages.
<details><summary>Click to expand. This screenshot shows that besides the first and last pages of this book, the other pages display 2 pages at a time</summary>

![Cdl Description of the Camera Lucida ID 7cee6c8f-0fd0-4f0c-88ed-8441fc4d16a7 Hyku Commons 2023-07-25 at 6 41 29 PM](https://github.com/scientist-softserv/palni-palci/assets/29311858/de5d553a-a29c-4285-9763-ffac8357e37b)
</details>

# Related
EPIC: #462

summer-cook:
### Softserv QA: Passed
[...]

PR #606, opened 2023-07-07
I600 uv customization

summer-cook:
# Story
Customize the UV for ReShare

# Related
- #600

- UV should not have buttons for download or share
- UV should have 2 page view enabled

# Screenshots / Video

Video showing 2 page view: 
https://share.getcloudapp.com/4guXlO1n

<details>
<summary>UV in ReShare Show Theme</summary>
image

</details>

Originally we were going to use an account setting for this, but since there are other customizations they want that didn't make a ton of sense anymore and it seemed better do it in the theme. See this for details. 
- #608

### DIFF
--- a/config/uv/uv-config-reshare.json
+++ b/config/uv/uv-config-reshare.json
@@ -1,0 +1,15 @@
+{
+  "modules": {
+    "footerPanel": {
+      "options": {
+        "downloadEnabled": false,
+        "shareEnabled": false
+      },
+      "contentLeftPanel": {
+        "options": {
+          "pageModeEnabled": true
+        }
+      }
+    }
+  }
+}
--- a/config/uv/uv-config-reshare.json
+++ b/config/uv/uv-config-reshare.json
@@ -4,11 +4,11 @@
       "options": {
         "downloadEnabled": false,
         "shareEnabled": false
-      },
-      "contentLeftPanel": {
-        "options": {
-          "pageModeEnabled": true
-        }
+      }
+    },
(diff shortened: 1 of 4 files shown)


config/uv/uv-config-reshare.json
```diff
@@ -0,0 +1,15 @@
+{
+  "modules": {
+    "footerPanel": {
+      "options": {
+        "downloadEnabled": false,
+        "shareEnabled": false
+      },
+      "contentLeftPanel": {
+        "options": {
+          "pageModeEnabled": true
```
labradford, line 10: I think the two page view is located in the header and not the left panel

summer-cook: yes good call, im just realizing its not working because my work with more than one page finally loaded. 

the line you had sent from nnp had it in this left panel thing, but 
theres also another pageModeEnabled in pagingHeaderPanel, but its set to false. should it be true?

https://github.com/scientist-softserv/nnp/blob/de9cf2427dff3ab6afebd9e3549b75bb5bd546c2/config/uv/uv-config.json#L196

config/uv/uv.html
```diff
@@ -16,31 +16,34 @@
                 margin: 0;
                 padding: 0;
                 overflow: hidden;
-            }
+              }
```
jeremyf, line 19: Reading notes: lines 19 to 22 are the only "substantive" change.  The others for this file are removing non-meaningful whitespace.

jeremyf (approved):

Issue #496, opened 2023-05-22
Create dynamic configuration for SSO


labradford:
# Story
Create a tab on the admin dashboard where you can enter SSO configuration options

Part of #472 
Part of #463 

# Acceptance criteria
- [ ] As an admin, I can visit a tab in the dashboard where I can enter SSO configuration options, either SAML or OIDC
- [ ] A dynamic form will load the appropriate fields for basic SAML or OIDC configurations

# Testing instructions

- [ ] To test on staging, log into https://reshare.palni-palci-staging.notch8.cloud/ as admin user
- [ ] Navigate to `Dashboard > Settings > Features`: https://reshare.palni-palci-staging.notch8.cloud/admin/features?locale=en
- [ ] Scroll down to `Show identity provider in admin dashboard`. Make sure this is enabled/turned on. This will enable the `Identity Provider` setting for you to test.
- [ ] Navigate to the `Dashboard > Settings > Identity Provider` to enter SSO Configuration.
- [ ] Choose the `Create New` button
  - [ ] Enter SSO configuration options
  - [ ] click the `Save` button
  - [ ] Verify options are saved
- [ ] To create new provider, click the delete button and create a new configuration
  - [ ] Enter SSO configuration options
  - [ ] click the `Save Changes` button
  - [ ] Verify options are saved

### The video tutorial is outdated so use the instructions above
<details><summary>Click to expand and view a video tutorial.</summary>

https://github.com/scientist-softserv/palni-palci/assets/29311858/5e9adb4d-eda6-40f5-923b-f2cd4494e7c3
</details>

# Notes
See code snippet generated by ChatGPT in #463 for a starting point.

This will require adding a new model and a new tab in the dashboard that is viewable by admin users only

We will be using this gem https://github.com/omniauth/omniauth-saml

Use documentation to decide basic SAML and OIDC fields

labradford:
```
config.omniauth :saml,
    allowed_clock_drift: 60,
    assertion_consumer_service_url: "https://#{Settings.multitenancy.root_host}/users/auth/saml/callback",
    attribute_statements: {
      email: ['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress'],
      name: ['http://schemas.xmlsoap.org/ws/2005/05/identity/claims/name']
    },
    certificate: File.read(Rails.root.join('provision/logingov.crt')),
[...]

PR #619, opened 2023-07-18
Rework for dynamic sso

summer-cook:
# Story
Auth providers were previously using the same properties to store their information, which was causing it to seem like the JS on the page was broken.

This pr gives each auth provider its own unique properties so that there is not confusing overlap and they change based on which provider the user chooses

It also fixes a bug that was discovered that allowed the user to create multiple auth providers

# Related
- #496 

- auth provider values would not always save as expected
- users could manually navigate to the `auth_providers/new` page and create multiple auth providers as many times as they wanted, which is not necessary. each tenant should have only one auth provider. 

- auth provider values save as expected
- switching between auth provider types will switch between their saved values in the form fields
- users can no longer navigate to the `auth_providers/new` page manually if they already have an existing provider saved. instead they will be redirected to the edit page of their existing provider (which will always be the first one) 

# Video
video: https://share.getcloudapp.com/OAul8lG0


summer-cook:
@labradford you will see in this PR video that the values for each provider are saved, so remain in the DB even if the user switched between oidc & saml.

If a user switched from one provider to another, should it wipe all the values for the other provider? Or is it fine the way it is/this is an edge case that won't happen very often anyways?

app/controllers/auth_providers_controller.rb
```diff
@@ -6,11 +6,21 @@ class AuthProvidersController < ApplicationController
 
   # GET /auth_providers/new
   def new
-    @auth_provider = AuthProvider.new
+    add_breadcrumbs
+    existing_auth_provider = AuthProvider.first
```
labradford, line 10: This might need a nil check

labradford (approved):

PR #646, opened 2023-08-04
Revert change and ignore rubocop

summer-cook:


app/services/hyrax/manifest_builder_service_decorator.rb
```diff
@@ -21,11 +21,13 @@ def loof(text)
         CGI.unescapeHTML(Loofah.fragment(text.to_s).scrub!(:prune).to_s)
       end
 
-      def sanitize_v3(input_hash:, presenter:, solr_doc_hits:)
-        hash = super(input_hash, presenter: presenter, solr_doc_hits: solr_doc_hits)
+      # rubocop:disable Lint/ShadowedArgument
+      def sanitize_v3(hash:, presenter:, solr_doc_hits:)
+        hash = super
```
jeremyf, line 26: I think this is not going to work.  The method signature of `sanitize_v3` is all key words, when you pass that to super it's going to be a method error.  Also, the shadow argument is something to consider and work to avoid, as it can become confusing what variable you're meaning.

jeremyf (changes requested):

app/services/hyrax/manifest_builder_service_decorator.rb
jeremyf: I read the IIIF Print definition/signature of `sanitize_v3`; it is `def sanitize_v3(hash:, presenter:, solr_doc_hits:)`.  So my point about parameters for super is mute.

However, I would like to consider renaming the local variable of `hash` to something other than `hash` (as that is the parameter name).  What about `returning_hash`?

jeremyf (approved):

Issue #626, opened 2023-07-20
Build the Platform report


labradford:

summer-cook:
for tenant pitt:

bundle exec rake counter_metrics:import_requests['pitt.hyku.test','spec/fixtures/csv/pittir-downloads.csv']

and 

bundle exec rake counter_metrics:import_investigations['pitt.hyku.test','spec/fixtures/csv/pittir-views.csv']

alishaevn:
`platform` is a required property of the `report_items` array and we're missing it. also, it looks like only `searches_platform` should be in the `performance` array. 

<details>
<summary>the docs</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/29032869/95ad990d-27b2-4860-b2e8-5702b562d5da)
</details>

<details>
<summary>our reports</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/29032869/56809aea-0862-4cc7-8ec5-1f271e68c060)
</details>

summer-cook:
We do need to add platform to report items, and we also need to bring the Platform data type over from the platform usage report. 

But, searches platform in the response example does contain total item investigations & total item requests for most data types, its just the platform data type that has searches platform. So I'm not sure if the docs are correct, or if the response if correct, but i think we should probably just believe the response since its probably better to give them those values since we already have them vs. not give them and them come back and want them. 

<img width="1023" alt="Screen Shot 2023-08-14 at 11 47 08" src="https://github.com/scientist-softserv/palni-palci/assets/73361970/2fc48518-4a67-49c8-8eef-f02bd2b55892">

alishaevn:
hmm, that is odd that the example response and the docs are contradictory. although I agree that since we've already but the extra fields in the response that it's easier to keep them. if pitt doesn't want them, then we can remove them.

PR #660, opened 2023-08-08
626 platform report

summer-cook:
# Story
Builds the counter compliant platform report

# Related
- #626 

- JSON returned a simple key/value pair

- [x] JSON for the platform report endpoint returns data for total investigations & total requests
- [x] Data can be filtered by begin date and end date
- [x] Data can be filtered by data_type (aligns with resource_type in hyku)

# Screenshots / Video

<details>
<summary>JSON</summary>
image

</details>


spec/models/platform_report_spec.rb
labradford, line 1: we should probably move this into a reports folder to follow the same path as the model file

summer-cook: good call

app/models/reports/platform_report.rb
```diff
@@ -0,0 +1,120 @@
...
+      @created = created
+      @account = account
+
+      # We want to limit the available attributes to be a subset of the given attributes; the `&` is
+      # the intersection of the two arrays.
+      @attributes_to_show = params.fetch(:attributes_to_show, ["Access_Method"]) & ALLOWED_REPORT_ATTRIBUTES_TO_SHOW
+
+      # Because we're receiving user input that is likely strings, we need to do some coercion.
+      @begin_date = self.class.coerce_to_date(params.fetch(:begin_date))
+      @end_date = self.class.coerce_to_date(params.fetch(:end_date))
```
labradford, line 43: The end data will always be the last day of the month, so coerce_to_date should take that into account.

labradford: probably can address this in a future PR

summer-cook: I think based on how it is set up, in that case they would query with begin date as the month they want, end date as the following month, or first day of the following month: for instance, begin_date=2023-06, end_date=2023-07 would be June 2023. Does that work?

labradford: The specification is begin date is first of the month, end date is the last day of the month. We can figure it out later

labradford (approved):

app/models/reports/platform_report.rb
alishaevn: @labradford for clarification:
- can we only search 1 month at a time? 
- or is it that we have to search in the span of an entire month? (e.g. all of august **and** september, but not aug 10 - sept 29)

alishaevn (approved):

app/models/reports/platform_report.rb
labradford: any reports will be for a complete month. there will be no partial month reports. For example, if you are querying for aug 2023, the response should be "that report is not ready" or something like that

PR #664, opened 2023-08-10
♻️ Renaming Sushi modules and fixing specs

jeremyf:
This commit contains a few changes:

1. Renaming Reports to Sushi
2. Moving the request specs into their module space
3. Fixing a broken test
4. Adding rudimentary error handling for missing parameters

In regards to renaming, my goal was to bring naming cohesion to the request spec, controller, and underlying models.  By setting things to all have Sushi as part of there name, this should make extraction and or mental cohesion of the concepts easier.

Regarding moving the request specs, adding the api folder seems like a good adjustment to help point out that these are part of an API (instead of being a top-level request spec).

Fixing the broken test speaks for itself.

And last, adding rudimentary error handling for missing parameters: this might not be the desired state, but it allows us to move forward with establishing how we handle missing required parameters.  One option is to fallback on defaults, another is to raise these kinds of errors.

If we provide defaults, we likely won't be raising exceptions, so the rescue block is mute.

As written in the comments, I don't want to rescue any old error and send that to the user; only errors in which they could provide meaningful adjustments to the request.


summer-cook (approved):

alishaevn:
I like it!

Issue #659, opened 2023-08-08
Allow tenant configuration to ignore showing PDFs in universal viewer


jeremyf:
## Summary

We want a feature flipper that turns on and off the PDF processing for iiif print.  This means a tenant can decide if they want PDFs showing in the viewer or not.  

## Testing instructions

Create a PDF work.  Go view it to see that it creates with a thumbnail.  
Go to the Settings, Features, and toggle the use iiif print feature.  
Upload a PDF and allow it to process.  
View the work and confirm you can see the PDF in the viewer.  

## Notes
There should be one single `flipflop` feature for the following tickets:

- #657
- #658 
- #659 

Testing Instructions can be found here:

- https://github.com/scientist-softserv/palni-palci/pull/666

The three tickets can all be tested in one swoop.

crisr15:
This fails SoftServ QA:  

- [ ] When I turn off the feature flipper for iiif print and upload a PDF, the file uploads properly and I do not see derivatives created, but the UV still displays an empty viewer.  
Example work: 
https://dev.commons-archive.org/concern/generic_works/4616118a-f8e4-4ace-bb9f-00f80e9073ae?locale=en

Screenshot: 

![Image](https://github.com/scientist-softserv/palni-palci/assets/11359350/b770b7c4-7ea2-48c8-b09b-8ee25471611c)

crisr15:
Passes internal QA.  Viewer does not show on works that do not have processed files.  



![Image](https://github.com/scientist-softserv/palni-palci/assets/11359350/1416342c-040f-4af2-8fbd-276faaa3dc32)

jeremyf:
Here's a thread reporting that this has been returned to SoftServ for further fixes: https://assaydepot.slack.com/archives/C0313NKC08L/p1692255870043969

Looking at the logs I'm seeing:

```
F, [2023-08-17T15:35:26.309458 #1] FATAL -- : [a0b4305bc8cff4fdb06cd6a1ffc94499] ActionView::Template::Error (private method `solr_document' called for #<Hyrax::IiifAv::IiifFileSetPresenter:0x00007f34682eab00>
Did you mean?  solr_document=):
F, [2023-08-17T15:35:26.309569 #1] FATAL -- : [a0b4305bc8cff4fdb06cd6a1ffc94499]     1: <%# Overridden from Hyrax 3.5.0 - To add extra download restrictions %>
[...]

Issue #658, opened 2023-08-08
Allow tenant configuration to ignore IIIF Print splitting PDFs consideration


jeremyf:
There should be one single `flipflop` feature for the following tickets:

- #657
- #658 
- #659 

Testing Instructions can be found here:

- https://github.com/scientist-softserv/palni-palci/pull/666

The three tickets can all be tested in one swoop.

crisr15:
Passes internal QA.  Not end user testable.

Issue #656, opened 2023-08-08
🦄 Spike: Determine How to Enable/Disable IIIF Prints Derivatives on a per Tenant Basis


jeremyf:
From [Slack Thread](https://assaydepot.slack.com/archives/C03B6A93E31/p1691440288645889?thread_ts=1691439565.576739&cid=C03B6A93E31): 

> Explore how we could do a feature flipper per tenant to turn off derivative creation from iiif print.  Because it adds to the storage and the speed of imports, we would like iiif print to be able to be turned off if a tenant does not want it.  So each tenant shoudl be able to select if pdf deriviatives are created in hyku.
>
> In order to get ReShare deployed we are going to need to deploy the hyku upgrade which includes iiif print to pals.  But we have some concerns at what it will do to storage requirements, upload time, and viewer load times, so we want hyku to have a feature flipper that allows each tenant to decide whether PDFs should show in viewer or not.  This will allow the community to have the iiif print and explore its uses, but not necessarily have everyone in a multi tenant environment using iiif print.

- [ ] #657
- [ ] #658
- [ ] #659

**_Note:_ Flipping the feature on will only impact processing in the future; to have the feature available to previously loaded work will require additional out of scope processes.  Likewise, disabling the feature will only impact future processing, there will still be derivatives and generated files from the "on" state.**

IIIF Print assumes certain SOLR fields; I suspect that these will not be relevant when the feature is disabled as those fields are used in parent/child relationships for viewer handling.

PR #666, opened 2023-08-10
🎁 Allowing per tenant ability to disable IIIF Print

jeremyf:
The current (as of <2023-08-10 Thu>) implementation of IIIF Print assumes that we generate derivatives at the application level.  However, we want options for generating at the tenant level.

As noted, I have chosen to default using IIIF print to `true` because that was the behavior before this commit.

In other words, this commit is functionally a non-change to the production code-base (assuming no one goes and flips some switches).

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/656
- https://github.com/scientist-softserv/palni-palci/issues/657
- https://github.com/scientist-softserv/palni-palci/issues/658
- https://github.com/scientist-softserv/palni-palci/issues/659

We always, regardless of configuration, use IIIF Print and therefore split PDFs and extract OCR.

Unless a tenant goes and turns off "use_iiif_print" there are no functional changes.

<details>
<summary>Screenshot of Flipflop dashboard with new feature</summary>
Screenshot of IIIF Print as flippable feature
</details>

# Verification

When they do disable "use_iiif_print" the way to verify that we're not splitting PDFs is to:

1. Create a new work with a PDF
2. Check that the work does not have child image objects attached (e.g. pages extracted from the PDF)


spec/services/iiif_print/tenant_config_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+
+    ##
+    # The purpose of this spec is to demonstrate that load sequence of modules correctly evaluates
+    describe '#service' do
+      subject { instance.service }
+
+      context 'when the feature is flipped to false' do
+        before { test_strategy.switch!(:use_iiif_print, false) }
+
+        it { is_expected.to eq(IiifPrint::SplitPdfs::ChildWorkCreationFromPdfService) }
```
laritakr, line 164: Doesn't false on the flipper mean you do NOT use IiifPrint, in which case the service would be the TenantConfig service?

spec/services/iiif_print/tenant_config_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      end
+
+      context 'when the feature is flipped to true' do
+        before { test_strategy.switch!(:use_iiif_print, true) }
+
+        it { is_expected.to eq(IiifPrint::TenantConfig::SkipSplittingPdfService) }
```
laritakr, line 170: See comment above re: false

laritakr (changes requested):
Very well thought out solution. Just two comments where if I followed everything, the specs seem to be reversed.

laritakr (approved):

alishaevn (approved):

Issue #627, opened 2023-07-20
Build the Item Report


labradford:
## Story
Part of #586 
Relies on #622 #616 Item Platform Report in JSON format. This report will be COUNTER compliant.

## Acceptance Criteria
- [ ] COUNTER compliant Item report is created
- [ ] Report is in JSON format
- [ ] Report uses historical data saved in the DB

## Testing
- Visit the route /api/sushi/r51/reports/ir and view JSON response for Platform Report
- Test suite should include tests for creation of the Item report

## Notes
More info on the Item report #595  

Historical Data can be imported using a rake task:
`bundle exec rake counter_metrics:import_investigations['pittir.hykucommons.org','spec/fixtures/csv/pittir-views.csv']`
and
`bundle exec rake counter_metrics:import_requests['pittir.hykucommons.org','spec/fixtures/csv/pittir-downloads.csv']`
*But use your local tenant cname

Works can be imported using csv file:

[pittir-import-works.csv](https://github.com/scientist-softserv/palni-palci/files/12285737/pittir-import-works.csv)

For the item report to be valid, you will need a full month of data. Use the above data as a starting point

summer-cook:
for me the commands were: 

`bundle exec rake counter_metrics:import_investigations['pitt.hyku.test','spec/fixtures/csv/pittir-views.csv']`
`bundle exec rake counter_metrics:import_requests['pitt.hyku.test','spec/fixtures/csv/pittir-downloads.csv']`

alishaevn:
updated the csv to have a `source_identifier` column instead of `eprintid`: 

[pittir-import-works.csv](https://github.com/scientist-softserv/palni-palci/files/12315394/pittir-import-works.csv)

alishaevn:
we need clarification about https://countermetrics.stoplight.io/docs/counter-sushi-api/5a6e9f5ddae3e-ir-item-report.

in the context of the docs above, but in relation to the data that will exist in the `Hyrax::CounterMetric` table, what is a parent item? what should be inside the `items` array? the `items` array is where the performance results are listed. however, you can have multiple items with no parent, a parent with 1 or more items and many options in between. 😕

alishaevn:
conversation about the above is happening in slack with clinton graham.
thread: https://assaydepot.slack.com/archives/C0313NKC08L/p1692036861164579

alishaevn:
# resources
[...]

PR #671, opened 2023-08-11
627 all items report

alishaevn:
# Story
create the items report

Refs
- #627 

- we are able to view all items at the `/reports/ir?<params>` endpoint
  - `/api/sushi/r51/reports/ir?begin_date=2023-06&end_date=2023-06`

# Screenshots / Video
image


app/models/sushi.rb
```diff
@@ -61,4 +61,79 @@ def coerce_data_types(params = {})
...
+            "Total_Item_Requests" =>
+            records.each_with_object({}) do |record, hash|
+              hash[record.year_month.strftime("%Y-%m")] = record.total_item_requests
+              hash
+            end
+          } }
+      end
+    end
+
+    def attribute_performance_for(resource_type:)
```
alishaevn, line 92: updated to accept a param so its reusable

app/models/sushi.rb
```diff
@@ -61,4 +61,79 @@ def coerce_data_types(params = {})
...
+                 .where("date >= ? AND date <= ?", begin_date, end_date)
+                 .order(:resource_type, "year_month")
+                 .group(:resource_type, "date_trunc('month', date)")
+
+      return relation if data_types.blank?
+
+      relation.where("LOWER(resource_type) IN (?)", data_types)
+    end
+
+    def data_for_resource
```
alishaevn, line 129: updated from `data_for_platform`

jeremyf (approved):

app/models/sushi.rb
jeremyf: I think this might introduce conditions for further introduction of a very subtle bug.  In adding the parameter, we're adjusting the label but we're not adjusting the SQL we're querying.  

The internal data for this method is very specifically a query for the platform.  With the method signature, it looks like I could give it "book" and expect to only have "book" results, even though we'd be querying the platform.

jeremyf (changes requested):

app/models/sushi.rb
alishaevn: taking a closer look at this made me realize I left some things out. marked back as a draft!

labradford (approved):

bkiahstroud (approved):

PR #675, opened 2023-08-11
🐛 Ensure consistent logic path for when iiif viewer

jeremyf:
There is a conflict between the `IiifPrint` and `Hyrax::IiifAv` gems; namely they both have strong opinions about how to sniff out if we should use the `iiif_viewer?`.  Compounding this, is that IiifPrint decorates `Hyrax::WorkShowPresenter` and Hyku extends `Hyrax::WorkShowPresenter` then includes
`Hyrax::IiifAv::DisplaysIiifAv`.

The end result is that the logic to determine if we should show pages split from the PDF is never called.  Yet, if we were to solely use IiifPrint we'd ignore rendering audio and vidoe in the iiif viewer.

So this commit peels that back so that we're using the logic (brought forward by IiifPrint) but ensuring our Hyku presenters are using that logic.

Why move the logic out of the Hyku instance and into a module?  Because that module contains the per-tenant antics of IiifPrint and its PDF relationship.

There is larger work to do in regards to incorporating this logic into IiifPrint and the Hyrax::IiifAv gem.

Does this work?  Please pull down a branch and check.  I'm at the end of my day (and then some).

Blocked for testing by:

- https://github.com/scientist-softserv/palni-palci/pull/691

### DIFF
--- a/app/presenters/hyku/work_show_presenter.rb
+++ b/app/presenters/hyku/work_show_presenter.rb
@@ -8,6 +8,13 @@ module Hyku
     # Hyrax::MemberPresenterFactory.file_presenter_class = Hyrax::FileSetPresenter
     # Adds behaviors for hyrax-iiif_av plugin.
     include Hyrax::IiifAv::DisplaysIiifAv
+
+    ##
+    # NOTE: IIIF Print prepends a IiifPrint::WorkShowPresenterDecorator to Hyrax::WorkShowPresenter
+    # However, with the above `include Hyrax::IiifAv::DisplaysIiifAv` we obliterate that logic.  So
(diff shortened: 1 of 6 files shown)


summer-cook (approved):

jeremyf:
To test this locally I:

- fetched the latest changes
- checked out this PR's branch
- cherry-picked 07071fa4161279e1675bad31a3472589d863057d (which comes from https://github.com/scientist-softserv/palni-palci/pull/691)
- Toggled off the IIIF Print flip flop flag and uploaded both images and PDF

bkiahstroud (approved):

Issue #686, opened 2023-08-14
Add filters to Platform Report


labradford:
# Story
Platform Reports should be able to receive filter params for the following:

- Begin_Date* required
- End_Date* required
- Platform(?)
- Access_Method (Regular or TDM - only regular users for now until we know if there are know data miners)
- Data_Type (this maps to the work resource_type)
- Metric_Type (Searches_Platform, Total_Item_Investigations, Total_Item_Requests, Unique_Item_Investigations, Unique_Item_Requests, Unique_Title_Investigations, Unique_Title_Requests)
- Exclude_Monthly_Details (true or false, default is false) - if false, we only send a total of all months in the date range and don't include individual months. The total should be included if this is set to true

**The following are listed in the SUSHI api docs but I do not see them in COUNTER. Not sure if they apply.**
- Attributed(?)
- Country_Code(?)
- Subdivision_Code(?)

# Acceptance Criteria

- [ ] Platform report can be filtered on the above items.

# Screenshots / Video

<details>
<summary></summary>

</details>

# Testing Instructions and Sample Files
**Developer to include URLs with the proper filters for testing**
- [ ] View the platform report in browser

# Notes

summer-cook:
**Already have:** 
Begin_Date
End_Date
Data_Type

**Need:**
Metric_Type
Exclude_Monthly_Details
Access_Method

**Probably don't need?**
Platform(?) - mostly sure we don't need this since the platform will always be the tenant where the reports are being made. they are already scoped by platform(tenant). You would visit the different tenants' base urls to see their reports. 
Attributed(?)
Country_Code(?)
Subdivision_Code(?)

summer-cook:
It seems like Exclude_Monthly_Details should not be used after all - it is only for tabular reports. The corresponding attribute for JSON reports is Granularity

See docs here: https://cop5.projectcounter.org/en/5.1/appendices/a-glossary-of-terms.html#appendix-a-glossary-of-terms
<img width="858" alt="image" src="https://github.com/scientist-softserv/palni-palci/assets/73361970/39f72fe6-72fb-4a57-af0a-723e92aae55a">

labradford:
QA - Platform Report filters are active and working

ctgraham:
# Pitt QA concerns:
## unexpected data
For a request filtered by `data_type` and `metric_type`, there is a `Attribute_Performance` with neither:
[...]

PR #693, opened 2023-08-15
686 filters - Metric Type filter only

summer-cook:
# Story
Sets up filtering by metric_type

# Related
Part of 
- #686 

- user could not filter by metric_type

- [ ] User can add `metric_type` to the params to filter which metric types are shown
- [ ] User can see their list of metric_types in the report filters under the "Metric_Type" key
- [ ] If no `metric_type` is added to the params, show all metric types (this is the default)
- [ ] Each report item will only show the metric types in the params, or the default (all types)
- [ ] Unique Title Investigations & Unique Title Requests only show for book datatypes
- [ ] If a non-allowed metric type is added, it does not appear in the params or the JSON
- [ ] the "Platform" report item still behaves differently, always appears and always shows the Searches Platform metric

# Screenshots / Video

Video: https://share.getcloudapp.com/7KuW9EJ6

<details>
<summary>JSON screenshot with filter</summary>
image

</details>

# Notes


spec/models/sushi/platform_report_spec.rb
```diff
@@ -4,31 +4,60 @@
...
+        {
+          begin_date: '2023-08',
+          end_date: '2023-09',
+          metric_type: 'total_item_investigations|unique_item_investigations|fake_value',
+          data_type: 'article',
+          attributes_to_show: ['Access_Method', 'Fake_Value']
+        }
+      end
+
+      it "only shows the requested metric types, , and does not include metric types that aren't allowed" do
```
alishaevn, line 45:
```suggestion
      it "only shows the requested metric types, and does not include metric types that aren't allowed" do
```

alishaevn (changes requested):

app/models/sushi/platform_report.rb
```diff
@@ -67,28 +71,19 @@ def attribute_performance_for_resource_types
...
-            end
-          } }
+          "Performance" => performance(records, resource_type) }
+      end
+    end
+
+    def performance(records, resource_type)
+      metric_types.each_with_object({}) do |metric_type, hash|
+        # Skip "Unique_Title_Requests" & "Unique_Title_Investigations" for all resource_types besides books
+        next if !resource_type.casecmp("book").zero? && (metric_type == "Unique_Title_Requests" || metric_type == "Unique_Title_Investigations")
```
jeremyf, line 81: I see that we're skipping everything but `"book"`, could you comment why that is the case?

jeremyf (changes requested):

app/models/sushi/platform_report.rb
jeremyf: > The Unique_Title metrics MUST NOT be used for Data_Types other than Book and Reference_Work as they are not meaningful for them. If a title contains both Open and Controlled sections or sections with different YOPs, the usage must be broken down by Access_Type and YOP so that the total counts are consistent between reports including and not including these columns/elements.

My read on the requirements is that we need to consider other resource_type; and I also suspect based on my read, that the unique may not be unique for a given day, but unique for a given person.

>  Unique_Title metrics were introduced in R5 to help normalize eBook metrics, and are retained in R5.1. Unique_Title metrics are only increased by 1 no matter how many (or how many times) chapters or sections are accessed in a given user session. Unique_Title metrics provide comparable eBook metrics regardless of the nature of the platform and how eBook content is delivered. They are comparable across report providers and across releases.

jeremyf (approved):

alishaevn (approved):

Issue #674, opened 2023-08-11
Why do banner images disappear after deploying Hyku upgrade


labradford:
# Story
After deploying the Hyku upgrade to production, banner images disappeared. Rolling back the deploy resolved the issue. 

The banner image loaded when Hyku upgrade was in production disappeared when the deploy was rolled back.

Did the file location for banner images change?

# Acceptance Criteria

- [ ] existing banner images do not disappear after deploy to production

# Screenshots / Video

<details>
<summary></summary>

</details>

# Testing Instructions and Sample Files

-

# Notes

alishaevn:
per leaann in standup: rolling back the deploy fixed this. so it may be a file paths problem.

aprilrieger:
I didnt see any branches reported or commits to point me to where to go, so I made an assumption that the branch currently deployed called production is the one we reverted to and the deploy that succeeded through actions but that cause the issue was production-deploy-8-10-23.

There were no apparent codes changes that occurred to cause the diff in behavior. I would need to conduct the deploy in real time to gather any further information. Or any logs you may have gathered during this issue.

labradford:
**before upgrade location:**
./public/uploads/f52d8231-8086-4c59-a71c-9f89d94e2396/site/banner_image/1/PXL_20210505_222328964.PORTRAIT.jpg

**after upgrade location:**
./public/system/banner_images/1/medium/bananas.jpg

Changed in this commit: https://github.com/samvera/hyku/commit/7ac76d7211f203853b8b3a392b184f1b12f58432

Switching to AvatarUploader

bkiahstroud:
### Further investigation findings

Copying/moving the banner images from the old location to the new expected location should work. However, if any of them have the same name, there will be a conflict and likely cause unexpected behavior / bugs. 

The old uploader class, [`Hyrax::UploadedFileUploader`](https://github.com/samvera/hyrax/blob/hyrax-v3.5.0/app/uploaders/hyrax/uploaded_file_uploader.rb), overrode the file storage location in a way that happened to play nicely with the multi-tenant nature of Hyku: the long alphanumeric in the middle of the path is the Tenant ID, ensuring each tenant will have its own unique banner image location:

```
# Old banner image file location
public/uploads/2b2726ff-19d5-48fc-ac18-8f47c0478245/site/banner_image/1/cat_scan.jpg
```

[...]

PR #702, opened 2023-08-16
i674 Rake tasks to move banner images to new expected file location

bkiahstroud:
# Story

Refs
- #674 

Adds two new rake tasks: 

- `hyku:copy_banner_images`
  - Will non-destructively copy all Hyku banner images across all tenants from the old file location to the new, expected location 
  - **Run this one first** 
- `hyku:delete_old_banner_images`
  - Clean up task: will delete all banner images in the old location 
  - **Run this one second _after_ confirming the presence of banner images across exiting tenants**

### DIFF
--- a/lib/tasks/move_banner_images.rake
+++ b/lib/tasks/move_banner_images.rake
@@ -27,7 +27,7 @@ namespace :hyku do
         # we can't simply point it to the new path
         Apartment::Tenant.switch(tenant_uuid) do
           site = Site.instance
-          File.open(renamed_destination) { |file| site.banner_image = file }
+          File.open(renamed_destination) { |banner_image| site.banner_image = banner_image }
           site.save!
         end
       else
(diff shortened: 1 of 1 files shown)


lib/tasks/move_banner_images.rake
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+# @see https://github.com/scientist-softserv/palni-palci/issues/674
+namespace :hyku do
+  desc 'Move all banner images to new expected location'
+  task copy_banner_images: :environment do
+    files_in_old_location = `find #{Rails.root}/public/uploads/*/site/banner_image/* -type f`.split("\n")
+    new_location = Rails.root.join('public', 'system', 'banner_images', '1', 'original')
```
jeremyf, line 8: Non-blocking, simply curious: what's the significance of the "1" directory?

bkiahstroud: That is the `Site` ID. Every tenant is represented by an `Account` object. Each `Account` has one `Site`. Since the `Site` db table is within the tenant-specific db schema, _every_ `Site` ID is guaranteed to be `1`.

bkiahstroud: I'll add a code comment explaining that.

labradford (approved):

PR #705, opened 2023-08-17
☑️  Add spring to provide option for faster tests

jeremyf:
To leverage spring, you will need to:

- Start the containers (e.g. `sc up`)
- Run the specs using spring (e.g. `spring rspec path/to/my_spec.rb`)

Spring boots up a Rails server and re-uses that between invocations. This means that instead of spending about 1 minute booting the application to run a 1 second test, you can re-use that app (spending about 8 seconds or so connecting to it) then run that 1 second test.

But beware of spring, as it can create odd behavior in the application. As a matter of practice when you are done testing, run `spring stop` to terminate the spring booted applications.

In past years, some of the pernicious bugs in development were because folks had spring running.  However, for unit tests, having a quick feedback is critical so I believe this is worth the risk.  Especially given that you must explicitly start spring.

Related to:

- https://github.com/scientist-softserv/britishlibrary/pull/252
- https://github.com/scientist-softserv/adventist-dl/pull/229

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -99,6 +99,7 @@ gem 'sidekiq'
 gem 'simplecov', require: false, group: %i[development test]
 gem 'solr_wrapper', '~> 2.0', group: %i[development test]
 gem 'spring', '~> 1.7', group: %i[development]
+gem 'spring-commands-rspec', group: %i[development]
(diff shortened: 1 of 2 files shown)


alishaevn (approved):
re-running test(3,2) to see if it'll pass.

Issue #688, opened 2023-08-14
Filtering for Item Report


labradford:
# Story
Item Reports should be able to receive filter params for the following:

- Begin_Date* required
- End_Date* required
- Platform(?)
- Author
- Item ID
- Access_Method (Regular or TDM - only regular users for now until we know if there are know data miners)
- Access Type (I think this is the item's visibility)
- Data_Type (this maps to the work resource_type)
- Metric_Type (Searches_Platform, Total_Item_Investigations, Total_Item_Requests, Unique_Item_Investigations, Unique_Item_Requests)
- YOP (Year of publish)

**The following are listed in the SUSHI api docs but I do not see them in COUNTER. Not sure if they apply.**
- Attributed(?)
- Country_Code(?)
- Subdivision_Code(?)

# Acceptance Criteria

- [ ] Item report can be filtered on the above items.

# Screenshots / Video

<details>
<summary></summary>

</details>

# Testing Instructions and Sample Files
**Developer to include URLs with the proper filters for testing**
- [ ] View the Item report in browser

# Notes

alishaevn:
blocked by #627

alishaevn:
https://countermetrics.stoplight.io/docs/counter-sushi-api/5a6e9f5ddae3e-ir-item-report
**access_type** string

Optional Report_Filter. A list of Access_Types separated by the | character (URL encoded as %7C) to return usage for. Omit this parameter to include all Access_Types.
Allowed values:
`Controlled` `Open` `Free_To_Read`


according to the above, I think we either need to not account for this filter at this time, or add it to the hyrax::countermetric datatable. if we do the latter, I'm not sure that the item visibility actually maps to the allowed values.
@labradford

alishaevn:
we discussed this in planning. below are notes from prior meetings about this filter. however, jeremy brought up good reasons as to why we may not want to account for this parameter at all.

although these methods don't 1:1 map to visibility, even if they did the visibility of an item can change. in that case, do the historical metrics change?? would all past metrics for the item have to be updated to reflect the current visibility? if the metrics didn't change historically, that also presents a problem. what if the visibility is private for the first 14 days of the month, then public for the rest of the days? reports are shown in 1 month increments, so which visibility status would be recorded? how would a use search properly for that?

<details>
<summary>previous notes</summary>

**What access types will you have? How would we determine what was what?**

[...]

PR #707, opened 2023-08-17
688 refactor query validation

alishaevn:
# Story
refactor the item report errors

instead of always throwing a 422 error for having no results, we will throw a 422 only if the parameter given is invalid. e.g., the item_id does not exist in the `Hyrax::CounterMetric` table.

if there are no results because of the date range given however, we will throw a 404 and inform the user of that.

- #688

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -1,9 +1,14 @@
 # frozen_string_literal:true
 
 module Sushi
-  class Sushi::NotFoundError < StandardError; end
-  class Sushi::MissingItemIdError < Sushi::NotFoundError; end
-  class Sushi::NoRecordsWithinDateRangeError < Sushi::NotFoundError; end
+  class Sushi::NotFoundError < StandardError
+    class << self
+      def no_records_within_date_range
+        new('There are no results for the given date range.')
+      end
+    end
+  end
+
 
   ##
   # Raised when the parameter we are given does not match our expectations; e.g. we can't convert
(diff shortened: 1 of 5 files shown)


app/models/sushi.rb
```diff
@@ -1,14 +1,23 @@
...
+
   ##
   # Raised when the parameter we are given does not match our expectations; e.g. we can't convert
   # the text value to a Date.
   class InvalidParameterValue < StandardError
     # rubocop:disable Metrics/LineLength
     class << self
       def invalid_item_id(item_id)
-        new("The given parameter `item_id=#{item_id}` did not return any results. Either there are no metrics for this id during the dates specified, or there are no metrics for this id at all. Please confirm all given parameters.")
+        new("The given parameter `item_id=#{item_id}` is invalid. Please provide a valid item_id, or none at all.")
```
jeremyf, line 20: This kind of exception should probably bubble up as a 404 error.  I'd rework this to also be a `Sushi::NotFoundError`.

Because they provided an item_id and that doesn't exist.  This maps to the default behavior of `ActiveRecord::NotFoundError`, in that when you request an ID that does not exist, it raises an `ActiveRecord::NotFoundError` but Rails by default rescues that to 404.

app/models/sushi.rb
```diff
@@ -187,7 +196,13 @@ def coerce_metric_types(params = {})
   module QueryParameterValidation
     extend ActiveSupport::Concern
     included do
-      attr_reader :platform
+      attr_reader :item_id, :platform
+    end
+
+    def validate_item_id(params)
+      raise Sushi::InvalidParameterValue.invalid_item_id(params[:item_id]) unless Hyrax::CounterMetric.exists?(['work_id LIKE ?', params[:item_id]])
```
jeremyf, line 203: In SQL `LIKE` is less performant than `=`.

Consider `exists?(work_id: params[:item_id])`

jeremyf: This should raise `Sushi::NotFoundError`

jeremyf (changes requested):

jeremyf (approved):

alishaevn:
`rspec ./spec/features/advanced_search_spec.rb:17` keeps failing in test (3,2,) but I'm unsure why. I've rerun the pipeline about 3-4 times now. the advanced search spec is unrelated to any changes in this pr.

Issue #687, opened 2023-08-14
Add filters to platform usage report


labradford:
# Story
Platform Usage Reports should be able to receive filter params for the following:

- Begin_Date* required
- End_Date* required
- Platform(?)
- Access_Method (Regular or TDM - only regular users for now until we know if there are know data miners)
- Metric_Type (Searches_Platform, Total_Item_Requests, Unique_Item_Requests, Unique_Title_Requests)

# Acceptance Criteria

- [ ] Platform report can be filtered on the above items.

# Screenshots / Video

<details>
<summary></summary>

</details>

# Testing Instructions and Sample Files
**Developer to include URLs with the proper filters for testing**
- [ ] View the platform report in browser

# Notes

labradford:
QA - Platform report is properly filtering

ctgraham:
# Pitt QA concerns:

## begin_date / end_date limits

When limiting the begin_date and end_date, metrics from outside of the range are still reported:
https://pittir.commons-archive.org/api/sushi/r51/reports/pr?begin_date=2023-06&end_date=2023-07&metric_type=Total_Item_Requests
Observed output:
```
"Total_Item_Requests":{"2023-05":98,"2023-06":700,"2023-07":740,"2023-08":521}
```
Expected output is:
```
"Total_Item_Requests":{"2023-06":700,"2023-07":740}
```

## begin_date after end_date

When the begin_date filter is after the end_date filter, a 3020 Exception is expected:
https://pittir.commons-archive.org/api/sushi/r51/reports/pr?begin_date=2023-04&end_date=2022-08&metric_type=Total_Item_Investigations
Observed output:
```
{"Report_Header":{"Release":"5.1","Report_ID":"PR","Report_Name":"Platform Report","Created":"2023-09-05T18:59:56Z","Created_By":"pittir.commons-archive.org","Institution_ID":{},"Institution_Name":"pittir.commons-archive.org","Registry_Record":"","Report_Attributes":{"Attributes_To_Show":["Access_Method"]},"Report_Filters":{"Begin_Date":"2023-04-01","End_Date":"2022-08-31","Metric_Type":["Total_Item_Investigations"]}},"Report_Items":{"Platform":"pittir.commons-archive.org","Attribute_Performance":[{"Data_Type":"Platform","Access_Method":"Regular","Performance":{"Searches_Platform":{}}}]}}
```
Expected output:
```
{
  "Code": 3020,
  "Message": "Invalid Date Arguments",
[...]

PR #710, opened 2023-08-17
687 Metric type filter for platform usage report

summer-cook:
# Story
Adds the metric type filter to the platform usage report

- #687 

- users could not filter by metric type

- [ ] As a user, i can add a metric_type filter to the params
- [ ] As a user, I should be able to filter by total_item_requests and unique_item_requests
- [ ] The platform usage report should only show requests  - not investigations, so if i add any investigation metric types, they should not appear in the JSON
- [ ] As a user, if i pass a random value to the metric_type filter param, nothing happens

# Screenshots / Video

<details>
<summary>JSON Structure when filtering by total_item_requests</summary>
image

</details>

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -163,20 +163,28 @@ module Sushi
       # "Unique_Title_Requests"
     ].freeze
 
-    def coerce_metric_types(params = {})
+    ALLOWED_METRIC_TYPES_PLATFORM_USAGE = [
+      # the platform usage report only contains requests. see https://countermetrics.stoplight.io/docs/counter-sushi-api/mgu8ibcbgrwe0-pr-p1-performance-other for details
+      "Unique_Item_Requests",
+      "Total_Item_Requests"
+    ].freeze
+
+    def coerce_metric_types(params = {}, platform_usage_report = false)
+      metric_types_array = platform_usage_report ? ALLOWED_METRIC_TYPES_PLATFORM_USAGE : ALLOWED_METRIC_TYPES
+
       metric_types_from_params = Array.wrap(params[:metric_type]&.split('|'))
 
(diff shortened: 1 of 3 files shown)


app/models/sushi.rb
```diff
@@ -163,20 +163,28 @@ module MetricTypeCoercion
...
     ].freeze
 
-    def coerce_metric_types(params = {})
+    ALLOWED_METRIC_TYPES_PLATFORM_USAGE = [
+      # the platform usage report only contains requests. see https://countermetrics.stoplight.io/docs/counter-sushi-api/mgu8ibcbgrwe0-pr-p1-performance-other for details
+      "Unique_Item_Requests",
+      "Total_Item_Requests"
+    ].freeze
+
+    def coerce_metric_types(params = {}, platform_usage_report = false)
```
jeremyf, line 172: The boolean for `platform_usage_report` is asking for trouble.  Namely we need to know what it means.

What if we were to instead adopt a different approach, namely adding a class method that returns the allowed metric types.  Then for the platform usage report, we overwrite that method.

I can pair on this to highlight the approach.

jeremyf (changes requested):
See inline comment

app/models/sushi/platform_usage_report.rb
```diff
@@ -48,25 +49,27 @@ def as_json(_options = {})
...
-            records.each_with_object({}) do |record, hash|
-              hash[record.year_month.strftime("%Y-%m")] = record.unique_item_requests
-              hash
-            end
-          } }
+          "Performance" => performance(records) }
+      end
+    end
+
+    def performance(records)
```
jeremyf, line 66: Non-blocking: I'm reading this and I think I understand but a future bit of work may want to look at removing the `next if`

jeremyf (approved):

PR #714, opened 2023-08-17
688 metric type filter

alishaevn:
# Story
allow for metric types to be filtered on
- #688 

# Screenshots / Video

TODO

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -158,33 +158,21 @@ module Sushi
     included do
       attr_reader :metric_types, :metric_type_in_params
     end
-    ALLOWED_METRIC_TYPES = [
-      "Total_Item_Investigations",
-      "Total_Item_Requests",
-      "Unique_Item_Investigations",
-      "Unique_Item_Requests",
-      # Unique_Title metrics exist to count how many chapters or sections are accessed for Book resource types in a given user session.
-      # This implementation currently does not support historical data from individual chapters/sections of Books,
-      # so these metrics will not be shown.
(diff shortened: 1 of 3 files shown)


app/models/sushi.rb
```diff
@@ -158,33 +158,21 @@ module MetricTypeCoercion
     included do
       attr_reader :metric_types, :metric_type_in_params
     end
-    ALLOWED_METRIC_TYPES = [
```
alishaevn, line 161: the list of allowed types differs per report (even if we aren't accounting for each of them right now). so I moved the const into the respective reports so`#coerce_metric_types` is still reusable.

app/models/sushi.rb
```diff
@@ -195,19 +183,22 @@ def coerce_metric_types(params = {})
   module QueryParameterValidation
     extend ActiveSupport::Concern
     included do
-      attr_reader :item_id, :platform
+      attr_reader :item_id, :item_id_in_params
+      attr_reader :platform, :platform_in_params
```
bkiahstroud, line 187: Nitpick (non-blocking): personally, this reads easier for me: 

```suggestion
      attr_reader :item_id, :item_id_in_params, :platform, :platform_in_params
```

Is there a reason to list them separately?

bkiahstroud (approved):

app/models/sushi.rb
alishaevn: I debated listing it the way you have it. I wondered if it may read easier to have the attributes grouped by likeness instead? I have more filters to do so I may change this in the future.

Issue #715, opened 2023-08-18
When IIIF Print is Disabled, We No Longer Render Representative Images (e.g. Thumbnail)


jeremyf:
As reported on <2023-08-17 Thu> by Nic:

> Post-update work pages with PDFs are no longer displaying the thumbnail of the file, instead displaying the default work icon or whatever image the user has set in place of the default icon.
>
> The download link normally present under the thumbnail is also missing. (edited)

# Code Analysis

The [Baseline Hyrax Implementation](https://github.com/samvera/hyrax/blob/b334e186e77691d7da8ed59ff27f091be1c2a700/app/views/hyrax/base/_representative_media.html.erb#L1-L9) is as follows:

```html
<% if presenter.representative_id.present? && presenter.representative_presenter.present? %>
  <% if defined?(viewer) && viewer %>
    <%= iiif_viewer_display presenter %>
  <% else %>
    <%= render media_display_partial(presenter.representative_presenter), file_set: presenter.representative_presenter  %>
  <% end %>
<% else %>
  <%= image_tag 'default.png', class: "canonical-image", alt: 'default representative image' %>
<% end %>
```

The [IIIF Print Override](https://github.com/scientist-softserv/iiif_print/blob/b170a57b24f63ca96873ca9b646d00f28d9c1b5f/app/views/hyrax/base/_representative_media.html.erb#L1-L10) is as follows:

```html
<% if presenter.iiif_viewer? %>
  <% if defined?(viewer) && viewer %>
    <%= iiif_viewer_display presenter %>
  <% else %>
    <%= render media_display_partial(presenter.representative_presenter), file_set: presenter.representative_presenter %>
  <% end %>
<% else %>
  <% alt = block_for(name: 'default_work_image_text') || 'Default work thumbnail' %>
  <%= image_tag default_work_image, class: "canonical-image", alt: alt %>
<% end %>
```

The logic has changed to say “If we are to use a IIIF Viewer…then render either the viewer OR the media\_display.”  In the case of toggling off IIIF Print for the tenant, the logic gates will presently result in showing a default work image.

# Story

# Acceptance Criteria

- [ ] When I have uploaded a PDF into a tenant that has disabled IIIF Print (which is now the default behavior), I should see the thumbnail of the work as the representative image.

# Proposed Solution

Ideally we should fix IIIF Print's overridden view.  However, for expediency and testing, we should copy the IIIF Print view into the repository and then amend the view.

My recommendation is as follows:

```html
[...]

PR #716, opened 2023-08-18
🐛 Render thumbnails when IIIF Print disabled

jeremyf:
Prior to this commit, the logic brough by IIIF Print was:

```html
<% if presenter.iiif_viewer? %>
  <% if defined?(viewer) && viewer %>
    <%= iiif_viewer_display presenter %>
  <% else %>
    <%= render media_display_partial(presenter.representative_presenter), file_set: presenter.representative_presenter %>
  <% end %>
<% else %>
  <% alt = block_for(name: 'default_work_image_text') || 'Default work thumbnail' %>
  <%= image_tag default_work_image, class: "canonical-image", alt: alt %>
<% end %>
```

We would check if we should use the IIIF viewer.  And because we'd disabled it for the tenant, the answer was no.  The fallback was to then render a default thumbnail.

Related to:

- https://github.com/scientist-softserv/iiif_print/issues/276
- https://github.com/scientist-softserv/palni-palci/issues/715
- https://github.com/scientist-softserv/iiif_print/pull/277

### DIFF
--- a/app/views/hyrax/base/_representative_media.html.erb
+++ b/app/views/hyrax/base/_representative_media.html.erb
@@ -1,0 +1,10 @@
+<% if presenter.representative_id.present? && presenter.representative_presenter.present? %>
+  <% if defined?(viewer) && viewer && presenter.iiif_viewer?%>
+    <%= iiif_viewer_display presenter %>
+  <% else %>
+    <%= render media_display_partial(presenter.representative_presenter), file_set: presenter.representative_presenter %>
+  <% end %>
+<% else %>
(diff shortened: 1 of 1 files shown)


alishaevn (commented):
looks good. waiting to approve until the pipeline is green.

alishaevn (approved):
approving since `spec/features/advanced_search_spec.rb:17` appears to be flaky.
ref: this [slack thread](https://assaydepot.slack.com/archives/C0311DNF3MG/p1692313076067139).

Issue #699, opened 2023-08-16
Make COUNTER metrics test data for staging


labradford:
# Story
Create a large dataset of works with counter metrics views and downloads for testing

# Acceptance Criteria

- [ ] rake task to make fake counter metrics for existing works in the repository
- [ ] there is 3 months worth of fake data
- [ ] optionally provide task with a list of known work ids

# Screenshots / Video

<details>
<summary></summary>

</details>

# Testing Instructions and Sample Files

- not end user testable

# Notes

labradford:
QA - running rake task in staging produced error `assign_attribute': unknown attribute 'year_of_publication' for Hyrax::CounterMetric`

summer-cook:
QA - was able to run this in staging. I needed to first run `rails db:migrate` and `bundle exec rake counter_metrics:generate_staging_metrics[dev.palni-palci-staging.notch8.cloud]`. I was able to generate the metrics for a single work. 

I have not been able to run it in demo yet in Pitt's testing env because it would not connect to the demo shell in rancher for some reason 🤷🏽‍♀️ But at least figured out that running migrations was indeed the issue

summer-cook:
Even though this is not end user testable, I'm going to leave this in the Softserv QA column for now until we can run it on demo & get the data into their staging environment for testing purposes.

labradford:
QA - ran the task on pittir.commons-archive.org successfully

PR #718, opened 2023-08-18
Rake task to get counter metric data on staging

summer-cook:
# Story
Sets up a rake task to get counter metric data on staging. 

# Related
- #699

- there was no real way to get counter metric data besides importing historical data. 

- [ ] As a developer, I can run `bundle exec rake counter_metrics:generate_staging_metrics[tenant_cname]` to generate a random number of `Hyrax::CounterMetric`s using random existing `GenericWorks`
- [ ] As a developer, I can run `bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test,'work_id_1|work_id_2|work_id_3']` to add `Hyrax::CounterMetric`s for a specified list of work ids. Work IDs must be passed in a string, separated by pipes.
- [ ] As a developer, I can run `LIMIT=1 bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test]` to generate a single counter metric. the limit can be changed according to the number of counter metrics i would like to generate. The default is 10

# Notes

### DIFF
--- a/app/services/generate_counter_metrics.rb
+++ b/app/services/generate_counter_metrics.rb
@@ -1,42 +1,36 @@
 # frozen_string_literal: true
 
-# Import counter views and downloads for a given tenant
-# Currently, this only imports the first resource type. Additional work would need to be done for works with multiple resource types.
-# The resource_type field in the counter_metrics table is single value and not an array.
+# Creates counter metrics for a single tenant.
 class GenerateCounterMetrics
-  def self.generate_counter_metrics(ids: 'all', limit: 10)
-    if ids == 'all' || ids.blank?
-      test_works = GenericWork.all.to_a.sample(limit)
+  def self.generate_counter_metrics(ids: :all, limit: :all)
+    fsq = "has_model_ssim: (#{Bulkrax.curation_concerns.join(' OR ')})"
+    fsq += " AND id:(\"" + Array.wrap(ids).join('" OR "') + "\")" if ids.present? && ids != :all
+    message = if ids == :all || ids.blank?
+                "Creating test data for #{limit} randomly selected works"
+              else
+                "Creating test data for works with ids #{ids}"
+              end
+    options = { fl: "id, has_model_ssim, resource_type_tesim", method: :post }
+    options[:rows] = limit if limit.is_a?(Numeric)
+    ActiveFedora::SolrService.query(fsq, options).each do |work|
+      work_type = work.fetch('has_model_ssim').first
+      work_id = work.fetch('id')
+      resource_type = work.fetch('resource_type_tesim').first
+      (0...90).each do |i|
+        date = i.days.ago.to_date
+        total_item_investigations = rand(1..10)
+        total_item_requests = rand(1..10)
 
-      test_works.concat(test_works.sample(limit - test_works.length)) while test_works.length < limit
-
-      message = "Creating test data for #{limit} randomly selected GenericWorks"
-    else
-      test_works = ActiveFedora::Base.where(id: ids).to_a
-      message = "Creating test data for works with ids #{ids}"
+        Hyrax::CounterMetric.create!(
+          worktype: work_type,
+          work_id: work_id,
+          resource_type: resource_type,
+          date: date,
+          total_item_investigations: total_item_investigations,
+          total_item_requests: total_item_requests
+        )
+      end
     end
     Rails.logger.info message
-    test_works.each do |work|
-      worktype = work.class
-      work_id = work.id
-      resource_type = work.resource_type.first
-      date = random_date_in_last_3_months
-      total_item_investigations = rand(1..10)
-      total_item_requests = rand(1..10)
-      Hyrax::CounterMetric.create!(
-        worktype: worktype,
-        work_id: work_id,
-        resource_type: resource_type,
-        date: date,
-        total_item_investigations: total_item_investigations,
-        total_item_requests: total_item_requests
-      )
-    end
-  end
-
-  def self.random_date_in_last_3_months
-    random_days_ago = rand(1..90) # 90 days in 3 months
-    random_date = Time.zone.today - random_days_ago
-    random_date.strftime("%Y%m%d")
   end
 end
--- a/app/services/generate_counter_metrics.rb
+++ b/app/services/generate_counter_metrics.rb
@@ -5,12 +5,19 @@
 # The resource_type field in the counter_metrics table is single value and not an array.
 class GenerateCounterMetrics
   def self.generate_counter_metrics(ids: 'all', limit: 10)
-    test_works = (ids == 'all' || ids.blank?) ?
-      GenericWork.all.to_a.sample(limit) :
-      ActiveFedora::Base.where(id: ids).to_a
-      ids.present? ?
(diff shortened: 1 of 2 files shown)


jeremyf (changes requested):
See inline comment

app/services/generate_counter_metrics.rb
```diff
@@ -0,0 +1,42 @@
...
+    else
+      test_works = ActiveFedora::Base.where(id: ids).to_a
+      message = "Creating test data for works with ids #{ids}"
+    end
+    Rails.logger.info message
+    test_works.each do |work|
+      worktype = work.class
+      work_id = work.id
+      resource_type = work.resource_type.first
+      date = random_date_in_last_3_months
```
jeremyf, line 23: Let's not assign a random date...let's have a date range (e.g. the last 3 months) that we loop through:

```ruby
(0..120).each do |integer|
  date =  integer.days.ago.to_date
  total_item_investigations = rand(1..10)
  total_item_requests = rand(1..10)
  Hyrax::CounterMetric.create!(
    ...
  )
```

lib/tasks/counter_metrics.rake
```diff
@@ -9,13 +9,35 @@ namespace :counter_metrics do
...
+  # bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test]
+  # Example with ids:
+  # bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test,'ab3c1f9d-684a-4c14-93b1-75586ec05f7a']
+  # Example with limit:
+  # LIMIT=1 bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test]
+  desc 'generate counter metric test data for staging'
+  task 'generate_staging_metrics', [:tenant_cname, :ids] => [:environment] do |_cmd, args|
+    raise ArgumentError, 'A tenant cname is required: `rake counter_metrics:generate_staging_metrics[tenant_cname]`' if args.tenant_cname.blank?
+    raise ArgumentError, "Tenant not found: #{args.tenant_cname}. Are you sure this is the correct tenant cname?" unless Account.where(cname: args.tenant_cname).first
+    ids = args.ids.present? ? args.ids.split("|") : []
```
jeremyf, line 34: If we don't have `args.id` I think we want to say "all".

app/services/generate_counter_metrics.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+# Creates counter metrics for a single tenant.
+class GenerateCounterMetrics
+  def self.generate_counter_metrics(ids: :all, limit: :all)
+    fsq = "has_model_ssim: (#{Bulkrax.curation_concerns.join(' OR ')})"
+    fsq += " AND id:(\"" + Array.wrap(ids).join('" OR "') + "\")" if ids.present? && ids != :all
+    message = if ids == :all || ids.blank?
+                "Creating test data for #{limit} randomly selected works"
```
jeremyf, line 9: I'm a bit confused by the message.  In particular why we're concerned about ids but then the message is about limits.

How about:

```ruby
message = "#{self.class}.generate_counter_metrics data for IDs = #{ids.inspect} with Limit = #{limit.inspect}."
```

Less conditionals and we're not worried about the interplay of `:ids` and `:limits` (which are mostly orthogonal in nature)

app/services/generate_counter_metrics.rb
```diff
@@ -0,0 +1,36 @@
...
+          worktype: work_type,
+          work_id: work_id,
+          resource_type: resource_type,
+          date: date,
+          total_item_investigations: total_item_investigations,
+          total_item_requests: total_item_requests
+        )
+      end
+    end
+    Rails.logger.info message
```
jeremyf, line 34: Let's move the `message` definition down to just above it's usage.

lib/tasks/counter_metrics.rake
```diff
@@ -9,13 +9,36 @@ namespace :counter_metrics do
...
     ImportCounterMetrics.import_investigations(args.csv_path)
   end
+
+  # You can optionally pass work IDs to have hyrax counter metrics created for specific works.
+  # Or, you can pass a limit for the number of CounterMetric entries you would like to create. currently they are randomly created.
+  # Example for pitt tenant in dev without ids:
+  # bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test]
+  # Example with ids:
+  # bundle exec rake "counter_metrics:generate_staging_metrics[pitt.hyku.test, ab3c1f9d-684a-4c14-93b1-75586ec05f7a|891u493hdfhiu939]"
+  # Example with limit of 1:
```
jeremyf, line 28:
```suggestion
  #
  # Example with limit of 1:
```

lib/tasks/counter_metrics.rake
```diff
@@ -9,13 +9,36 @@ namespace :counter_metrics do
...
     raise ArgumentError, 'A tenant cname is required: `rake counter_metrics:import_investigations[tenant_cname]`' if args.tenant_cname.blank?
     raise ArgumentError, "Tenant not found: #{args.tenant_cname}. Are you sure this is the correct tenant cname?" unless Account.where(cname: args.tenant_cname).first
     AccountElevator.switch!(args.tenant_cname)
     puts "Importing historical investigations for #{args.tenant_cname}"
     ImportCounterMetrics.import_investigations(args.csv_path)
   end
+
+  # You can optionally pass work IDs to have hyrax counter metrics created for specific works.
+  # Or, you can pass a limit for the number of CounterMetric entries you would like to create. currently they are randomly created.
+  # Example for pitt tenant in dev without ids:
```
jeremyf, line 24:
```suggestion
  #
  # Example for pitt tenant in dev without ids:
```

lib/tasks/counter_metrics.rake
```diff
@@ -9,13 +9,36 @@ namespace :counter_metrics do
...
+  # bundle exec rake counter_metrics:generate_staging_metrics[pitt.hyku.test]
+  # Example with ids:
```
jeremyf, line 26:
```suggestion
  #
  # Example with ids:
```

lib/tasks/counter_metrics.rake
jeremyf: The three suggestions give some breathing room to the examples; helps scalability of code.

lib/tasks/counter_metrics.rake
```diff
@@ -9,13 +9,36 @@ namespace :counter_metrics do
     ImportCounterMetrics.import_requests(args.csv_path)
   end
 
-  # bundle exec rake counter_metrics:import_requests['pittir.hykucommons.org','spec/fixtures/csv/pittir-downloads.csv']
   desc 'import historical counter investigations'
+  # bundle exec rake counter_metrics:import_requests['pittir.hykucommons.org','spec/fixtures/csv/pittir-downloads.csv']
```
jeremyf, line 13: Would it make sense to move the example comments up to this area as well?

I also think comments should happen before the `desc` declaration. 

```ruby
##
# inline comment
desc "Terse description"
task ...
```

jeremyf (approved):

PR #719, opened 2023-08-21
688 access method filter

alishaevn:
# Story
create validations for the `access_method` query param.

if the `access_method` query param is passed in with at least one valid option, we return the valid option(s) to
 be reflected in the `report filters` array.
if no valid options are passed through the param, we throw a `Sushi::InvalidParameterValue` error.

either way, the `access_method` that shows in the `item_report` is hard coded to say `Regular` because that's the only option we are allowing at this time. if that changes, the code in Sushi::ItemReport#report_items will need to be updated.

- #688 

# Screenshots / Video
image

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -195,6 +195,7 @@ module Sushi
     ##
     # @param params [Hash, ActionController::Parameters]
     def validate_access_method(params)
+      return true unless params.key?(:access_method)
       allowed_access_methods_from_params = Array.wrap(params[:access_method].split('|')).map { |am| am.strip.downcase } & ALLOWED_ACCESS_METHODS
 
       raise Sushi::InvalidParameterValue.invalid_access_method(params[:access_method], ALLOWED_ACCESS_METHODS) unless allowed_access_methods_from_params.any?
@@ -206,6 +207,7 @@ module Sushi
     ##
     # @param params [Hash, ActionController::Parameters]
     def validate_item_id(params)
+      return true unless params.key?(:item_id)
       raise Sushi::NotFoundError.invalid_item_id(params[:item_id]) unless Hyrax::CounterMetric.exists?(work_id: params[:item_id])
 
       @item_id = params[:item_id]
@@ -215,6 +217,7 @@ module Sushi
     ##
     # @param params [Hash, ActionController::Parameters]
     def validate_platform(params, account)
+      return true unless params.key?(:platform)
       raise Sushi::InvalidParameterValue.invalid_platform(params[:platform], account) unless params[:platform] == account.cname
 
       @platform = account.cname
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -22,6 +22,10 @@ module Sushi
       def invalid_platform(platform, account)
         new("The given parameter `platform=#{platform}` is not supported at this endpoint. Please use #{account.cname} instead. (Or do not pass the parameter at all, which will default to #{account.cname})}")
       end
+
+      def invalid_access_method(access_method, acceptable_params)
+        new("None of the given values in `access_method=#{access_method}` are supported at this time. Please use an acceptable value, (#{acceptable_params.join(', ')}) instead. (Or do not pass the parameter at all, which will default to the acceptable value(s))")
+      end
     end
     # rubocop:enable Metrics/LineLength
   end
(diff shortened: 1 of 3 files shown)


app/models/sushi.rb
```diff
@@ -183,22 +187,38 @@ def coerce_metric_types(params = {}, allowed_types: ALLOWED_METRIC_TYPES)
...
 
       @item_id = params[:item_id]
-      @item_id_in_params = params.key?(:item_id)
+      @item_id_in_params = true
     end
 
+    ##
+    # @param params [Hash, ActionController::Parameters]
     def validate_platform(params, account)
-      raise Sushi::InvalidParameterValue.invalid_platform(params[:platform], account) unless params[:platform].blank? || params[:platform] == account.cname
```
alishaevn, line 198: the `params[:platform].blank?` was written before we were checking that the param key exists before calling this method.

technically a user could write `platform=` and this exception wouldn't be raised, but I don't think that's a use case we have to account for.

app/models/sushi/item_report.rb
```diff
@@ -73,10 +74,11 @@ def as_json(_options = {})
 
       raise Sushi::NotFoundError.no_records_within_date_range if report_items.blank?
 
-      report_hash['Report_Header']['Report_Filters']['Item_ID'] = item_id if item_id_in_params
-      report_hash['Report_Header']['Report_Filters']['Platform'] = platform if platform_in_params
+      report_hash['Report_Header']['Report_Filters']['Access_Method'] = access_methods if access_method_in_params
       report_hash['Report_Header']['Report_Filters']['Data_Type'] = data_types if data_type_in_params
+      report_hash['Report_Header']['Report_Filters']['Item_ID'] = item_id if item_id_in_params
       report_hash['Report_Header']['Report_Filters']['Metric_Type'] = metric_types if metric_type_in_params
+      report_hash['Report_Header']['Report_Filters']['Platform'] = platform if platform_in_params
```
alishaevn, line 81: alphabetized.

app/models/sushi/item_report.rb
```diff
@@ -45,8 +45,9 @@ def initialize(params = {}, created: Time.zone.now, account:)
 
     def validate_item_report_parameters(params:, account:)
       coerce_metric_types(params, allowed_types: ALLOWED_METRIC_TYPES)
-      validate_item_id(params) if params[:item_id]
-      validate_platform(params, account) if params[:platform]
+      validate_access_method(params) if params.key?(:access_method)
```
jeremyf, line 48: Hmm…there's a coupling of knowledge between the validate methods and the conditionals here.

Namely, we are assuming knowledge of the hash key here and inside the validate methods.  What if the validate methods were to instead introduce the guard clause?

Here we would have:

```ruby
validate_item_id(params)
```

Then elsewhere

```ruby
def validate_item_id(params = {})
  return true unless params.key?(:item_id)
  ...
```

Then there's reduced duplication of knowledge regarding the key that we're using for the `item_id`

jeremyf (changes requested):

app/models/sushi/item_report.rb
alishaevn: I considered adding a guard clause too....but then wondered if it would be "better" to skip entering the method at all, vs calling it and potentially immediately exiting. but I do like your point about the duplication of knowledge.

jeremyf: Another thing to consider.  If I chose to locally override the method, would I want the knowledge of the guard clause within that method or outside of the method?  My preference is within.

Also, method calls are very inexpensive compared to our cognitive load of reading the initialize method, then holding that "if" statement while we go and read the validation method (remembering the "if" statement)

app/models/sushi.rb
```diff
@@ -183,22 +187,38 @@ def coerce_metric_types(params = {}, allowed_types: ALLOWED_METRIC_TYPES)
...
+    end
+
+    ##
+    # @param params [Hash, ActionController::Parameters]
     def validate_item_id(params)
       raise Sushi::NotFoundError.invalid_item_id(params[:item_id]) unless Hyrax::CounterMetric.exists?(work_id: params[:item_id])
 
       @item_id = params[:item_id]
-      @item_id_in_params = params.key?(:item_id)
+      @item_id_in_params = true
```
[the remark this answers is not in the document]
alishaevn: with [f1b7d90](https://github.com/scientist-softserv/palni-palci/pull/719/commits/f1b7d909c573f176adbe1e56ee52e38582e9aabf) we're checking for the presence of the param in each validation method. which still means `<xyz>_in_params` will always be true.

jeremyf (approved):

PR #722, opened 2023-08-21
♻️ Reduce # of Rows to Process for Sushi

jeremyf:
Based on #701, in reducing the number of rows we process, we generate
the report faster.

Related to:

- https://github.com/scientist-softserv/palni-palci/pull/701


alishaevn (approved):
that `advanced_search_spec` failed again. 🙄 approving anyway since we know it's trippy. (I'm re-running the spec just in case though)

Issue #584, opened 2023-06-27
Set up Shibboleth SAML Auth


labradford:

labradford:
- Link to SAML is added to login page
- clicking link redirects to Pitt Passport page with this error:

![Image](https://github.com/scientist-softserv/palni-palci/assets/18175797/cd8fa966-cabc-4e3c-87aa-fea9b77e5478)

URL request:
```
fZFLa8MwEIT%2Fim862XKch4OIU0xCIZA%2BiNseeimys8YisuRqV2nz76s09AGF3vYw8%2B3s7AJlrwdReurMDl49IEUlIjhS1qysQd%2BDq8AdVQOPu23BOqIBBed7OCbd6eATChbugwO5DBR%2BBvJGal3L5nClbRihAMOidRAqI8%2FgH8wgEQfrKBkUUQJ7z9V%2B4IOzrdLAq07VtdUQsFV1x6LNumAv%2BaidjNoxxPW0nsWTPM%2FjeV1DPB3XkELbTGfZJEgRPWwMkjRUsCzNxnE6i7P8IRuLUSqm82cWPYXQn2myJGXRe68NinP8gnlnhJWoUBjZAwpqRFXebEUQCvlVz2%2FL8L8nHES2sZotF5fCb4Nks763WjWnqNTavq0cSIKCkfPAomvrekkXaC8bEPhdhRgF4HnFZg%2BGVKvAMb5c8L%2BfXH4A
```



decodes to 

```
<samlp:AuthnRequest 

AssertionConsumerServiceURL='https://dev.hyku.test/users/auth/saml/callback?locale=en' 

Destination='https://passport.pitt.edu/idp/profile/Shibboleth/SSO' ID='_71f41f3e-b5b6-4777-8bbe-53be0efc5624' 

IssueInstant='2023-06-27T23:10:58Z' 

Version='2.0' xmlns:saml='urn:oasis:names:tc:SAML:2.0:assertion' xmlns:samlp='urn:oasis:names:tc:SAML:2.0:protocol'>

<samlp:NameIDPolicy 

AllowCreate='true' 

Format='urn:mace:shibboleth:1.0:nameIdentifier'/>

</samlp:AuthnRequest>
```

labradford:
[...]

PR #726, opened 2023-08-21
I584 Pitt SSO use display name instead of generated email

labradford:
# Story

Refs #584 
When logging in to Pitt via SSO, the username was a long string on nonsense instead of the user's display name. This PR adds Pitt specific code that will generate the display name from the response.

User's display name was long string + @example.com
User's display name will be the users first + last name

# Screenshots / Video

Screenshot 2023-08-22 at 10 49 03 AM

# Notes


summer-cook (approved):

app/models/user.rb
```diff
@@ -35,7 +35,10 @@ def self.from_omniauth(auth)
     find_or_create_by(provider: auth.provider, uid: auth.uid) do |user|
```
jeremyf, line 35: This does not look like it resolves their username/email?

labradford:
close in favor of https://github.com/scientist-softserv/palni-palci/pull/756

Issue #720, opened 2023-08-21
YOP report filter


alishaevn:
# Story
update the following reports to respond to the `yop` query param.
https://countermetrics.stoplight.io/docs/counter-sushi-api/yc1o09env0d2a-counter-sushi-api

- [ ] Item report

# Acceptance Criteria

- [ ] Item report can be filtered by the `yop`.

# Screenshots / Video

# Testing Instructions and Sample Files
**Developer to include URLs with the proper filters for testing**
- [ ] View the Item report in browser

alishaevn:
blocked by #695

jeremyf:
Do we know know what Fedora property contains the year of publication?

jeremyf:
Given that we can have multiple values:  `1999|2003-2005|2007` we are going to want to write a something to convert those into an array of integers (e.g. `[1999, 2003, 2004, 2005, 2007]`).  That splitter would be a good Sushi module method.

alishaevn:
for additional reference:

<details>
<summary>slack thread</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/29032869/137b3e61-8a92-4752-b262-7628ff58037a)
</details>

alishaevn:
this ticket is still blocked by #695. the column was added to the `hyrax_counter_metrics` table, but the`ImportCounterMetrics` class hasn't been updated to use it.

alishaevn:
https://pittir.commons-archive.org/api/sushi/r51/reports/ir?begin_date=2023-06&end_date=2023-06&yop=2002

![Image](https://github.com/scientist-softserv/palni-palci/assets/29032869/e75e5e06-fdb9-45be-89a4-96743032f758)

ctgraham:
# Pitt QA concerns:
## YOP report filter in the header

A successful YOP filter should appear as an element in the `Report_Filters` element in the `Report_Header`.
https://pittir.commons-archive.org/api/sushi/r51/reports/ir?begin_date=2023-06&end_date=2023-06&yop=2000-2020
Observed:
```
"Report_Header": {
    "Report_Name": "Item Report",
    "Report_ID": "IR",
    "Release": "5.1",
    "Institution_Name": "pittir.commons-archive.org",
    "Institution_ID": {},
    "Report_Filters": {
      "Begin_Date": "2023-06-01",
      "End_Date": "2023-06-30"
    },
    "Created": "2023-09-05T19:39:07Z",
[...]

PR #758, opened 2023-08-23
🎁 Adding year_of_publication filter to Sushi report

jeremyf:
Before this commit, we were not handling the `:yop` query parameter.

With this commit, we're parsing the given `:yop` to build the date range
query over the `Hyrax::CounterMetric#year_of_publication` field.

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/720
- https://github.com/scientist-softserv/palni-palci/issues/688

### DIFF
--- a/spec/models/sushi_spec.rb
+++ b/spec/models/sushi_spec.rb
@@ -142,4 +142,54 @@ RSpec.describe Sushi do
       it { is_expected.to be_nil }
     end
   end
+
+  describe "YearOfPublicationCoercion" do
+    subject(:instance) { klass.new(params) }
+
+    let(:klass) do
+      Class.new do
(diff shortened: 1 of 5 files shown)


summer-cook (approved):

spec/models/sushi_spec.rb
```diff
@@ -142,4 +142,54 @@
...
+    subject(:instance) { klass.new(params) }
+
+    let(:klass) do
+      Class.new do
+        include Sushi::YearOfPublicationCoercion
+        def initialize(params = {})
+          coerce_yop(params)
+        end
+      end
+    end
```
alishaevn, line 156: I tried to write up a spec for `QueryParameterValidation` in here last week but couldn't get it to work. I think something like this is what I was missing.

Issue #633, opened 2023-07-27
ReShare - Set up a work lease for authenticated user


labradford:

labradford:
Per client we won't be getting a time stamp but will require users to re-login after 24 hours

jeremyf:
I started looking at the UCSD and there is the idea of WorkAuthorizations; this partially manages the authorizations by creating records in the database for a work and user and their authorization token; that token is then used to interact with an API to manage the state of the authorization.

I’m not seeing the term “lease” in the code base.

jeremyf:
Here is the desired local workflow:

- [ ] As adminsitrator create two works (A and B), make both of them private.  You’ll want to save the URL for both of these works.
- [ ] Create a new user (do not grant them access to either work)
- [ ] Logout
- [ ] Go to the URL of work A.  You will be redirected to sign-in.
- [ ] Sign in as the non-admin user
  - You should be able to see work A.
- [ ] Go to work B.  You should be redirected to sign-in.
- [ ] Sign in as the non-admin user
  - You should be able to see work B
- [ ] Go to the work A, you should be able to see work A without being prompted for authentication.

One challenge:  When you are already logged in and go to a page that is private (e.g. you can’t see) the system notes that this is private and does not prompt you to authenticate.

I'm thinking about creating a development only service that is configured for after authentication to grant access to the work at the original URL.  This would allow for continued local testing and help us move forward without all of the SSO work.

jeremyf:
I have the following Devise hack for localized testing:

```ruby
# This decorator splices into Devise to manage the authorization expiration
module DeviseSessionDecorator
  def create
    # We have authenticated the user, but have not yet issued the respond_with (e.g. redirect to the
    # stored location).
    super do
      # Note: Accessing `stored_location_for` clears that location from the session; so we need to
      # grab it and then set it again.
      url = stored_location_for(:user)
      store_location_for(:user, url)

      work_pid = work_pid_from_url(url)
      WorkAuthorization.handle_signin_for!(user: current_user, work_pid: work_pid)
    end
  end

  private

  REGEXP_TO_MATCH_PID = %r{/concern/[^\/]+/(?<work_pid>[^\?]+)(\?.*)?}.freeze
  def work_pid_from_url(url)
    return unless url.present?
    match = REGEXP_TO_MATCH_PID.match(url)
    return unless match
    match[:work_pid]
  end
end

# Don't unleash this into production code; however the functionality is useful for testing.
[...]

PR #766, opened 2023-08-31
🎁 Add AuthN/AuthZ for CDL assets

jeremyf:
## 🎁 Add AuthN/AuthZ for CDL assets

58a5728d34193c6521258f1f6537a4bc701f315c

Prior to this commit, we had not wired in the logic for authorization to
controlled digital lending (CDL) assets.  With this change, we're adding
the logic for handling a user authenticating to access a work.

The logic flow is as follows:

An unauthenticated user goes to a CDL asset page.  They are neither
authorized nor authenticated, so we direct them to the new session
route.  There we capture we're the URL of the requested resource and
send them on their way to authenticate.

When we perform the handshake with OpenID, we provide (as scope and
client design decision) the URL of the CDL asset.  The OpenID provider
authenticates and returns to the OmniAuth callback.  We use the returned
scope to then grant authorization to the CDL asset (and revoke any
expired authorizations).

And once they login for the CDL asset, we redirect to that asset.

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/633
- https://github.com/scientist-softserv/palni-palci/pull/647

## handle csrf issue with openid state. use the redirect pattern all the time since its hapazard which sso vendors need it

f39e0d38831dd216cef6872d94d81aabff3b20a7

## ♻️ Appease rubocop and fix parameter mismatch

0ed1777ab2a8c504c67798541db6372e81732076

## Appease rubocop

663ae641e3e5d08c3c08a5eaf5bf5ad3e867d23f

## ♻️ Specify redirect URL for CDL SSO

6de21f6021be6db66fc41efc4f4cf0b8cfaccf37

Prior to this commit, we rendered a page that used JS to change the
window's location.

With this commit, we pass to the rendered page the URL we want to use
for redirection.

## 🎁 Ensure non-admin user friendly routing for SSO

6c9a3eda0506cc073239d2613d04822d4bbde46b

We are trying to serve two types of users:

- Admins
- Not-admins

Given that admins are a small subset, we can train and document how they
can sign in.  In other words, favor workflows that impact the less
trained folk to help them accomplish their tasks.

Prior to this commit, given the site had an SSO provider, when I (an
unauthenticated user) went to a private work, then it would redirect me
to the `/user/sign_in` route.

At that route I had the following option:

1. Providing a username and password
2. Selecting one of the SSO providers to use for sign-in.

The problem with this behavior was that a user who was given a
Controlled Digital Lending (CDL) URL would see a username/password and
likely attempt to authenticate with their CDL username/password (which
was managed by the SSO provider).

The end result is that the authentication page most likely would create
confusion.

With this commit, I'm setting things up such that when the application
uses calls `new_user_session_path` we make a decision on what URL to
resolve.

Related to:

- https://github.com/scientist-softserv/palni-palci/pull/766
- https://github.com/scientist-softserv/palni-palci/pull/647
- https://github.com/scientist-softserv/palni-palci/issues/633

## 🐛 Ensure proper order for URL in scope

199268feae2fe979eeddec1206ca39b486021a7e

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -23,14 +23,9 @@ module Users
 
         # We need to render a loading page here just to set the sesion properly
         # otherwise the logged in session gets lost during the redirect
-        if params[:action] == 'saml'
-          set_flash_message(:notice, :success, kind: params[:action]) if is_navigational_format?
-          sign_in @user, event: :authentication # this will throw if @user is not activated
-          render 'complete'
-        else
-          sign_in_and_redirect @user, event: :authentication # this will throw if @user is not activated
-          set_flash_message(:notice, :success, kind: params[:action]) if is_navigational_format?
-        end
+        set_flash_message(:notice, :success, kind: params[:action]) if is_navigational_format?
+        sign_in @user, event: :authentication # this will throw if @user is not activated
+        render 'complete'
       else
         session['devise.user_attributes'] = @user.attributes
         redirect_to new_user_registration_url
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -5,11 +5,22 @@ module Users
     skip_before_action :verify_authenticity_token
 
     def callback
-      logger.info("========== auth: #{request.env['omniauth.auth']}")
+      logger.info("=@=@=@=@  auth: #{request.env['omniauth.auth']}, params: #{params.inspect}")
 
       @user = User.from_omniauth(request.env['omniauth.auth'])
 
       if @user.persisted?
+        # https://github.com/scientist-softserv/palni-palci/issues/633
+        WorkAuthorization.handle_signin_for!(user: @user, scope: params[:scope])
+
+        # By default the sign_in_and_redirect method will look for a stored_location_for.  However,
+        # we're seeing that the stored location seems to get lost.  So we'll rely on the presence of
+        # the scope to handle this.
+        #
+        # Related to OmniAuth::Strategies::OpenIDConnectDecorator
+        url = WorkAuthorization.url_from(scope: params[:scope], request: request)
+        store_location_for(:user, url) if url
+
         # We need to render a loading page here just to set the sesion properly
         # otherwise the logged in session gets lost during the redirect
         if params[:action] == 'saml'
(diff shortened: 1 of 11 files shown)


app/helpers/hyrax_helper.rb
```diff
@@ -4,6 +4,7 @@ module HyraxHelper
   include ::BlacklightHelper
   include Hyrax::BlacklightOverride
   include Hyrax::HyraxHelperBehavior
+  prepend IiifPrint::IiifHelperDecorator
```
orangewolf, line 7: are you lost little guy? do you need help finding your mommy or daddy pr?

app/controllers/users/omniauth_callbacks_controller.rb
```diff
@@ -10,6 +10,9 @@ def callback
       @user = User.from_omniauth(request.env['omniauth.auth'])
 
       if @user.persisted?
+        # https://github.com/scientist-softserv/palni-palci/issues/633
+        WorkAuthorization.handle_signin_for!(user: @user, scope: params[:scope])
```
orangewolf, line 14: I think this needs a conditional? will this break on non-params scope calls? remember all other SSO calls come through this same method

app/helpers/hyrax_helper.rb
orangewolf: not a big deal - just making sure this should be here

orangewolf (changes requested):

app/helpers/hyrax_helper.rb
jeremyf: This little guy did sneak in.  And I'll work on removing it.

app/controllers/users/omniauth_callbacks_controller.rb
jeremyf: The logic of `WorkAuthorization.handle_signin_for!` uses the user logging in as the time to add/revoke authorization to works.  If `params[:scope]` is `nil`, then we won't grant authorization to any new works.  And we'll revoke any work authorizations that have expired.

```ruby
  def self.handle_signin_for!(user:, authorize_until: 1.day.from_now, work_pid: nil, scope: nil, revoke_expirations_before: Time.zone.now)
    pids = Array.wrap(work_pid) + extract_pids_from(scope: scope)

    # Maybe we get multiple pids; let's handle that accordingly
    pids.each do |pid|
      begin
	authorize!(user: user, work_pid: pid, expires_at: authorize_until)
      rescue WorkNotFoundError
	Rails.logger.info("Unable to find work_pid of #{pid.inspect}.")
      end
    end

    # We re-authorized the above pids, so it should not be in this query.
    where("user_id = :user_id AND expires_at <= :expires_at", user_id: user.id, expires_at: revoke_expirations_before).pluck(:work_pid).each do |pid|
      revoke!(user: user, work_pid: pid)
    end
  end
```

orangewolf (approved):

Issue #125, opened 2022-11-08
Embed PDF viewer


orangewolf:
# Summary
Implement pdf.js viewer: https://mozilla.github.io/pdf.js/

An embedded PDF viewer is an essential improvement for the implementation and use of Hyku by both our current and potential user base. They would like to have the option to use IIIF Print or the embedded PDF for PDF items through feature flippers. 

# Acceptance Criteria
- [x] https://mozilla.github.io/pdf.js/ is an embedded PDF viewer
- [x] Indocument search is set up
- [x] Near instantaneous file loading
- [x] Configuration option to display full PDF in vertical format
- [x] Full page display at 100% scale
- [x] Single page display with “next” and “previous” buttons for readers to move between pages
- [x] Ability for readers to jump to indicated pages
- [x] Select and cut options for text
- [x] Feature flipper for embedded PDF is available for PDF items
- [x] Feature flipper for IIIF Print is available for PDF items
- [x] Instructions that enforce only one feature flipper to be enabled are displayed with the above feature flippers

# Screenshots or Video
Example of non-cached pdf loading in JSTOR:

![Image](https://user-images.githubusercontent.com/29311858/200966679-49b7ae44-3e55-4bca-a87c-17a4b0489704.png)

Example of vertical format:

![Image](https://user-images.githubusercontent.com/29311858/200966692-26007532-2ec2-4188-aab5-90ffbadf2308.png)

# Testing Instructions
- Create a work with a PDF attached to it
- Default Theme
  - [ ] Indocument search is set up
  - [ ] Near instantaneous file loading
  - [ ] Configuration option to display full PDF in vertical format
  - [ ] Full page display at 100% scale
  - [ ] Single page display with “next” and “previous” buttons for readers to move between pages
  - [ ] Ability for readers to jump to indicated pages
  - [ ] Select and cut options for text
  - [ ] Feature flipper for embedded PDF is available for PDF items
  - [ ] Feature flipper for IIIF Print is available for PDF items
  - [ ] Instructions that enforce only one feature flipper to be enabled are displayed with the above feature flippers
- Cultural Theme
  - [ ] Indocument search is set up
  - [ ] Near instantaneous file loading
  - [ ] Configuration option to display full PDF in vertical format
  - [ ] Full page display at 100% scale
  - [ ] Single page display with “next” and “previous” buttons for readers to move between pages
  - [ ] Ability for readers to jump to indicated pages
  - [ ] Select and cut options for text
[...]

PR #769, opened 2023-09-06
🎁 Install PDF.js viewer

kirkkwang:
# Story

Install PDF.js as the default viewer for PDF resources.  This will only work for works that only have a PDF as a file set.  A feature flipper has been added to modify the use of PDF.js or UV as the default PDF viewer.  Switching this to UV will leverage the features of IIIF Print to split PDFs into images and run Tesseract OCR on them and store the text to be retrieved by the IIIF Search API.  Switching the default PDF viewer to UV will require re-ingesting of the PDF so they can run through the IIIF Print process.

Refs #125 

- UV was the default viewer for PDFs

- PDF.js is the default viewer for PDFs
- There is a feature flipper to indicate which PDF viewer is set as the default

# Screenshots / Video

image

image


ShanaLMoore (approved):
great job Kirk! it looks good. :)

ShanaLMoore:
These specs are flaky. Locally it passes: 
image
image

Issue #721, opened 2023-08-21
author report filter


alishaevn:
# Story
update the following reports to respond to the `author` query param.
https://countermetrics.stoplight.io/docs/counter-sushi-api/yc1o09env0d2a-counter-sushi-api

- [ ] Item report

# Acceptance Criteria

- [ ] Item report can be filtered by the `author`.

# Screenshots / Video

# Testing Instructions and Sample Files
**Developer to include URLs with the proper filters for testing**
- [ ] View the Item report in browser

alishaevn:
blocked by #695

jeremyf:
Do we know what Fedora properties we should count as `author`?

alishaevn:
> Do we know what Fedora properties we should count as author?

@jeremyf my understanding is that we are pointing it at `creator`.
ref: [slack thread](https://assaydepot.slack.com/archives/C05HUG89YVD/p1692631643844159?thread_ts=1692137277.543219&cid=C05HUG89YVD).

alishaevn:
for additional reference:

<details>
<summary>another slack thread</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/29032869/137b3e61-8a92-4752-b262-7628ff58037a)
</details>

alishaevn:
this ticket is still blocked by #695. the column was added to the `hyrax_counter_metrics` table, but the`ImportCounterMetrics` class hasn't been updated to use it.

summer-cook:
@labradford I think that is because none of the data in the pittir.commons-archive.org have an author with that name (i just searched, and it seems like every item has an author of "null" in that tenant?)

I will check it on another tenant that has some authors

summer-cook:
hmmm ok, just tested it again at 

https://dev.palni-palci-staging.notch8.cloud/api/sushi/r51/reports/ir?begin_date=05-2023&end_date=08-202&author=Summer

there is one work in the db with a creator of "Summer" haha. it did not find that one work

alishaevn:
> QA - getting this error on pittir.commons-archive.org "error": "The given parameter `author=Hamilton, Carol V.` was malformed. Please provide the first author's name exactly as it appears on the work."

[...]

PR #771, opened 2023-09-06
🎁 Add multiple author handling to Sushi report

jeremyf:
Prior to this commit, we made the short-cut assumption that each work would have one author.

With this commit, we now serialize authors in the `author` field and then query accordingly.  The query is setup to be backwards compatible with already imported data.

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/721

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -288,6 +288,36 @@ module Sushi
     end
 
     ##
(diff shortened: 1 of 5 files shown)


alishaevn (approved):
approving, but the pipeline needs to be fixed.

Issue #772, opened 2023-09-08
Full text search capability for text, including OCR text—when searching within Hyku, search results should include the text of embedded pdfs


jillpe:
# Summary
Pals would like OCR text, and full text search capability for searching text from a PDF through the catalog

# Accepted Criteria
- [x] A PDF can be found by inputting it's text in the search catalogue
- [ ] When found in the search results, a text snippet displays with it's search result

# Testing Instruction
- Create a work with a PDF and make sure the viewer is visible

In-Viewer Search
- Click the search icon at the top of the viewer
- Search a term
  - [ ] the term is identified and highlighted on the PDF wherever it occurs
  
Catalogue Search
- in the catalogue search, search for a term that is in the PDF
  - [ ] the PDF shows as a search result and the term is highlighted in the text snippet that appears with it

jillpe:
SoftServ QA: 

[PDF tested](https://dev.commons-archive.org/concern/generic_works/47a56d40-58f7-4de1-94bd-769f5f2c67ea?locale=en)

<details><summary>In-viewer search ✅ </summary>


![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/81cfe6f7-3f88-4481-918e-ffd625a6b0c5)


</details>


<details><summary>Catalogue Search ❌ </summary>
no search results found



![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/9e970b84-b438-4270-a0c7-a08cc3ce00da)


</details>

PR #781, opened 2023-09-11
:gift: Enable snippet highlighting for search

ShanaLMoore:
In this PR we are indexing the pdf text onto the parent record, adding a highlight to the matching search results and displaying the snippet in the search catalog. An index will be required when we deploy to the various environments. 

# Story

Refs #772 

Search results were not displaying as a highlight on the search catalog page. 

Screenshot 2023-09-11 at 3 39 55 PM

### DIFF
--- a/app/assets/stylesheets/hyrax.scss
+++ b/app/assets/stylesheets/hyrax.scss
@@ -20,3 +20,9 @@
 @import "blacklight_gallery/osd_viewer";
 
 @import "themes/*";
+
+// for catalog search result snippets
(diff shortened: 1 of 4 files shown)


kirkkwang (approved):
Looks good to me!  I wish there was a better place for the css class though but i don't think it's that big of a deal

ShanaLMoore:
> Looks good to me! I wish there was a better place for the css class though but i don't think it's that big of a deal

We can make our own css file and include it in application css

PR #786, opened 2023-09-12
🐛 PDF.js looks better in other themes

kirkkwang:
Originally this was developed on only the default theme and the other themes looked weird.  This commit will fix that.

# Story

Refs #125 

image

image


ShanaLMoore (approved):
Just noting that we could clean this up by making a method vs littering `@presenter.file_set_presenters.any?(&:pdf?)` in all the views.

PR #787, opened 2023-09-12
🐛 Account for IIIF Print logic

kirkkwang:
This commit will guard the work indexing with a Flipflop check so we don't get stuck at the AttachFilesToWorkJob.

# Story

Refs #772 

If an admin switched to UV which uses IIIF Print logic, the works never get completely created.

The works complete.

### DIFF
--- a/app/indexers/app_indexer.rb
+++ b/app/indexers/app_indexer.rb
@@ -21,7 +21,7 @@ class AppIndexer < Hyrax::WorkIndexer
   end
 
   def full_text(file_set_id)
-    return if file_set_id.blank?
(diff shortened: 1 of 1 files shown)


app/indexers/app_indexer.rb
```diff
@@ -21,7 +21,7 @@ def generate_solr_document
   end
 
   def full_text(file_set_id)
-    return if file_set_id.blank?
+    return if !Flipflop.default_pdf_viewer? || file_set_id.blank?
```
ShanaLMoore, line 24: currently snippets is only available for PDF.js

ShanaLMoore (approved):

PR #793, opened 2023-09-15
🐛 Remove PDF.js related fields from form

kirkkwang:
The `show_pdf_viewer` and `show_pdf_download_button` fields were showing up on the form so this commit will remove them.

# Story

Refs #issuenumber

image

image

### DIFF
--- a/app/forms/hyrax/pdf_form_behavior.rb
+++ b/app/forms/hyrax/pdf_form_behavior.rb
@@ -5,7 +5,14 @@ module Hyrax
     extend ActiveSupport::Concern
 
     included do
+      class_attribute :hidden_terms
+
       self.terms += %i[show_pdf_viewer show_pdf_download_button]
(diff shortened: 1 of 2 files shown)


app/forms/hyrax/pdf_form_behavior.rb
```diff
@@ -5,7 +5,14 @@ module PdfFormBehavior
     extend ActiveSupport::Concern
 
     included do
+      class_attribute :hidden_terms
+
       self.terms += %i[show_pdf_viewer show_pdf_download_button]
+      self.hidden_terms = %i[show_pdf_viewer show_pdf_download_button]
```
ShanaLMoore, line 11: I like your solution and think it'd be a helpful one to contribute back to hyku even, but I'm going to see if i can figure out why I can't just not include it in the form's terms.

ShanaLMoore (approved):

PR #795, opened 2023-09-15
🐛 Fix double render error

kirkkwang:
Sometimes when you search for a phrase the results will throw an exception.  This commit will fix that by calling the IIIF Print version of the `#render_ocr_snippets` method.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/772


ShanaLMoore (approved):
inspired by a similar problem that LaRita fixed: https://github.com/scientist-softserv/adventist-dl/pull/475/files

Issue #743, opened 2023-08-22
By default sort featured collection works by A-Z 


jillpe:
## Summary
Collections are alphabetical order

![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/b2c5083d-04d8-4ee0-8f91-6740df82f61b)

# NOTE

**Unfortunately, because this ticket's screenshot was misleading, we implemented default A-Z sort on the featured collection section instead of the collection's show page.**

**Use case, when a user clicks "feature" on collections, by default it will be displayed in alphabetical order. If a user manually drags to a different order, it will persist to how they've organized it.**

**I created a new ticket that matches the original intent of this ticket: https://github.com/scientist-softserv/palni-palci/issues/800**

The original title for this ticket said: "By default sort collection works by A-Z on collection page rather than relevance"

## Accepted Criteria

- [ ] ~Works on the collection page are by default, sorted A-Z~
- [ ] When a user clicks "feature" on collections, by default it will be displayed in alphabetical order. 
- [ ] If a user manually drags to a different order, it will persist to how they've organized it.

ShanaLMoore:
@aahurford 

> Note -- this should relate to the collection page, not the home page. We can currently control the sort on the home page -- we want to change the default works sort in collection pages to A-Z rather than relevance. 
> 
> https://goshen.hykucommons.org/collections/c0c98979-9b5a-41b4-95bb-606af556e30e

Because the original ticket as misleading, we implemented default A-Z sort on the featured collection section. Should we undo this work? 

Use case, when a user clicks "feature" on collections, by default it will be displayed in alphabetical order. If a user manually drags to a different order, it will persist to how they've organized it.

I created a new ticket that matches your screenshot: https://github.com/scientist-softserv/palni-palci/issues/800

kirkkwang:
A feature of two collections

<img width="749" alt="Screenshot 2023-09-18 at 08 45 14" src="https://github.com/scientist-softserv/palni-palci/assets/19597776/2167fb97-b4a3-4097-a790-324798691f04">

Then add another, note how they are sorted alphabetically by default instead of by last added

<img width="780" alt="Screenshot 2023-09-18 at 08 45 55" src="https://github.com/scientist-softserv/palni-palci/assets/19597776/e9305c9e-2e20-4526-b95f-ba9ae1eb1ba9">

ShanaLMoore:
[...]

PR #796, opened 2023-09-15
🎁 Default A-Z title sort for featured collections

kirkkwang:
This commit will add a default A-Z title sort to the featured collections section on the homepage.  If a featured collections list has been saved then it will no long sort by title but instead whatever the user set it to.

# Story

Refs #743 

By default, featured collections would be sorted by the last added.

By default, featured collections would be sorted by title.

# Screenshots / Video

https://github.com/scientist-softserv/palni-palci/assets/19597776/04bb3109-ec5e-4de4-9e6b-b60510bbb605

# Notes
When the user saves a list, this functionality will stop working until the list has been cleared and new collections added.

### DIFF
--- a/app/models/featured_collection_list.rb
+++ b/app/models/featured_collection_list.rb
@@ -22,6 +22,8 @@ class FeaturedCollectionList
       collection.destroy if collection.presenter.blank?
       collection.presenter.blank?
     end
+    sort_by_title! unless manually_ordered?
+    @collections
   end
 
   delegate :empty?, to: :featured_collections
@@ -47,5 +49,13 @@ class FeaturedCollectionList
 
     def collection_with_id(id)
       @collections.find { |c| c.collection_id == id }
+    end
+
+    def manually_ordered?
+      !@collections.all? { |c| c.order == FeaturedCollection::FEATURE_LIMIT }
+    end
+
+    def sort_by_title!
+      @collections.sort_by! { |c| c.presenter.title.first.downcase }
     end
 end
--- a/spec/models/featured_collection_list_spec.rb
+++ b/spec/models/featured_collection_list_spec.rb
@@ -3,16 +3,20 @@
 require 'rails_helper'
 
 RSpec.describe FeaturedCollectionList, type: :model do
+  subject { described_class.new }
+
   let(:user) { create(:user).tap { |u| u.add_role(:admin, Site.instance) } }
   let(:account) { create(:account) }
   let(:collection1) { create(:collection, user: user) }
   let(:collection2) { create(:collection, user: user) }
+  let(:featured_collection_1) { create(:featured_collection, collection_id: collection1.id) }
+  let(:featured_collection_2) { create(:featured_collection, collection_id: collection2.id) }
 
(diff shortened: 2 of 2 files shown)


app/models/featured_collection_list.rb
```diff
@@ -22,6 +22,8 @@ def featured_collections
       collection.destroy if collection.presenter.blank?
       collection.presenter.blank?
     end
+    sort_by_title! unless manually_ordered?
```
ShanaLMoore, line 25: If there are specs for #featured_collection this may be a good time to add a context to it.

kirkkwang: added specs!

ShanaLMoore (approved):

ShanaLMoore (changes requested):
approved after some lint fixes

app/models/featured_collection_list.rb
ShanaLMoore: thanks it looks great!

spec/models/featured_collection_list_spec.rb
```diff
@@ -3,16 +3,17 @@
 require 'rails_helper'
 
 RSpec.describe FeaturedCollectionList, type: :model do
+  subject { described_class.new }
   let(:user) { create(:user).tap { |u| u.add_role(:admin, Site.instance) } }
   let(:account) { create(:account) }
   let(:collection1) { create(:collection, user: user) }
   let(:collection2) { create(:collection, user: user) }
+  let!(:featured_collection_1) { create(:featured_collection, collection_id: collection1.id) }
```
ShanaLMoore, line 11: reviewing the lint errors, one of them is about the use of Let!. 

A few ideas... because it doesn't look like you are using these variables, you could move this into a before block instead. 

```ruby
  # let!(:featured_collection_1) { create(:featured_collection, collection_id: collection1.id) }
  # let!(:featured_collection_2) { create(:featured_collection, collection_id: collection2.id) }
  
  before do
    create(:featured_collection, collection_id: collection1.id)
    create(:featured_collection, collection_id: collection2.id)
  end
```

Otherwise you would have to call on the variable as the first step within your it block to activate it before your expectation.

spec/models/featured_collection_list_spec.rb
```diff
@@ -35,6 +36,20 @@
         expect(presenter.id).to eq collection2.id
       end
     end
+
+    context 'when the featured collections have not been manually ordered' do
+      it 'is sorted by title' do
+        allow(subject).to receive(:manually_ordered?).and_return(false)
```
ShanaLMoore, line 42: Another lint errors on stubbing subjects. Looking closer at this test, I don't think you will need to stub it at all if you implement either suggestion from above. 

Another semi related idea, you could test whether or not each context receives :sort_by_title

kirkkwang: i incorporated the suggestion and removed the `let!` but still needed to somehow control what `#manually_sorted` returned, please take a look

ShanaLMoore (approved):

PR #764, opened 2023-08-24
sushi-dates-filtering

alishaevn:
# Story
account for the available dates in the `Hyrax::CounterMetric` table when searching by begin_date and end_data query params

# Expected Behavior Before Changes

# Expected Behavior After Changes

# Screenshots / Video

<details>
<summary></summary>

</details>

# Notes

PR #804, opened 2023-09-19
Resolving sushi error handling

jeremyf:
The [spec defines numerous exceptions][1].  This refactor moves towards
conforming to those specifications.

This change also builds on raising exceptions when data is not yet
ready.  To then guard against those exceptions, we need to either rescue
in the JSON document (which would be rather gross) or wrap the Sushi
methods with a guarding function.  I chose the latter.

See the inline comments for further discussion/documentation on these
changes.

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/688

[1]: https://countermetrics.stoplight.io/docs/counter-sushi-api/sgmqulr0emsi6-exception

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -1,51 +1,6 @@
 # frozen_string_literal:true
 
 module Sushi
-  class NotFoundError < StandardError
-    class << self
-      def no_records_within_date_range
-        new('There are no results for the given date range.')
-      end
-
-      def invalid_item_id(item_id)
-        new("The given parameter `item_id=#{item_id}` is invalid. Please provide a valid item_id, or none at all.")
-      end
-    end
-  end
-
-  ##
-  # Raised when the parameter we are given does not match our expectations; e.g. we can't convert
-  # the text value to a Date.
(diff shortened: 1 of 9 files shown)


app/models/sushi.rb
```diff
@@ -151,6 +133,10 @@ def coerce_dates(params = {})
       # Because we're receiving user input that is likely strings, we need to do some coercion.
       @begin_date = Sushi.coerce_to_date(params.fetch(:begin_date)).beginning_of_month
       @end_date = Sushi.coerce_to_date(params.fetch(:end_date)).end_of_month
+
```
alishaevn, line 136: since we're here...can we also account for when the begin or end date fall outside of the dates we have data for? (ref: #764)

alishaevn (changes requested):

app/models/sushi.rb
jeremyf: This is a significant and chunky change to the specs; I have the written production logic stenciled in but it seriously breaks a lot of specs.

My strong preference would be to merge this and then build on that requirement.

alishaevn (approved):

Issue #814, opened 2023-09-22
Missing Download Button


jillpe:
## Summary
https://assaydepot.slack.com/archives/C0313NKC08L/p1695393021454599

[Example](https://palsave.hykucommons.org/concern/parent/b3eed969-aac8-4312-b20e-b9f90ef3b574/file_sets/eae14c1a-a7dd-4a69-8b06-c4e63a6892b0)

## Accepted Criteria
- [ ] Download pdf button displays and works on file pages

PR #788, opened 2023-09-13
I773 configurable display

ShanaLMoore:
# Story

Refs #773 

# Expected Behavior After Changes

Admins should be able to choose if a work should render the PDF.js viewer or the Download PDF button.

# Screenshots / Video

https://github.com/scientist-softserv/palni-palci/assets/19597776/892b4a95-98f6-4f06-88ee-6494378f85f7

# Notes
Currently the defaults are to show the viewer and the download button.  However, a reindex is required for this behavior.

PR #815, opened 2023-09-22
🐛 Bring back missing download link on pdf

kirkkwang:
This commit will fix a bug where the Download PDF link was erroneously removed in a previous commit.  It was originally removed because it was rendering on the work show page but in doing so it was also removed from the file set show page.  We now bring it back with a conditional that only renders it on the file set show page.

Ref
  - https://github.com/scientist-softserv/palni-palci/issues/814
  - https://github.com/scientist-softserv/palni-palci/pull/788

image

### DIFF
--- a/app/views/hyrax/file_sets/media_display/_pdf.html.erb
+++ b/app/views/hyrax/file_sets/media_display/_pdf.html.erb
@@ -6,6 +6,13 @@
                     class: "representative-media",
                     alt: "",
                     role: "presentation" %>
+      <% if controller_name == 'file_sets' %>
+        <%= link_to t('hyrax.file_set.show.downloadable_content.pdf_link'),
+              hyrax.download_path(file_set),
+              target: :_blank,
+              id: "file_download",
+              data: { label: file_set.id } %>
+      <% end %>
     </div>
 <% else %>
     <div>


ShanaLMoore (approved):

app/views/hyrax/file_sets/media_display/_pdf.html.erb
```diff
@@ -6,6 +6,13 @@
                     class: "representative-media",
                     alt: "",
                     role: "presentation" %>
+      <% if controller_name == 'file_sets' %>
```
jeremyf, line 9: Do we need logic of something akin to `can?(:download, file_set)`?

ShanaLMoore: it's not really related to ability though. maybe a method to remove this logic from the view will suffice?

kirkkwang: Good question, it wasn't there prior and plus a few lines above we have this:

```erb
<% if display_media_download_link?(file_set: file_set) && (Site.account.settings[:allow_downloads].nil? || Site.account.settings[:allow_downloads].to_i.nonzero?) %>
```

jeremyf: If it wasn't there prior, then that's good for me!

jeremyf (approved):

PR #816, opened 2023-09-22
🐛 Handle `nil` passing to Markdown

kirkkwang:
This commit will fix a bug relating to the ocr snippets.  Because we are using the the `markdown` helper, we need to make sure we are  not passing `nil` to it otherwise it'll break.  We are getting nil because we don't find any snippets and IIIF Print would return a nil so we don't display empty quotes on the search page.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -38,6 +38,7 @@ module ApplicationHelper
       disable_indented_code_blocks strikethrough lax_spacing space_after_headers
       quote footnotes highlight underline
     ]
+    text ||= ""
     Markdown.new(text, *options).to_html.html_safe
   end
 end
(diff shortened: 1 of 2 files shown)


jeremyf (commented):
It looks like https://github.com/scientist-softserv/palni-palci/blob/2ecbdf6536c537202901b6044a8fe8e250e5c5fe/app/helpers/application_helper.rb#L35-L42 defines the `markdown` method, what about going in there and having 

```ruby
  def markdown(text)
    options = %i[
      hard_wrap autolink no_intra_emphasis tables fenced_code_blocks
      disable_indented_code_blocks strikethrough lax_spacing space_after_headers
      quote footnotes highlight underline
    ]
    text ||= ""
    Markdown.new(text, *options).to_html.html_safe
  end
```

laritakr (approved):

PR #821, opened 2023-09-25
688 exception 3050

alishaevn:
# Story
validate the parameters given to the reports
- #688 

- you could pass an invalid parameter and the report would render

- if you pass an invalid parameter, a 3050 exception error will be rendered

# Screenshots / Video

image

# Notes

### DIFF
--- a/app/models/sushi.rb
+++ b/app/models/sushi.rb
@@ -113,6 +113,23 @@ module Sushi
   end
 
   ##
+  # @param params [Hash, ActionController::Parameters]
+  # @param allowed_parameters [Array<String>]
+  #
+  # @return [String]
+  # @raise [Sushi::Error::UnrecognizedParameterError] when any disallowed params are given.
+  module ParameterValidation
+    def validate_paramaters(params = {}, allowed_parameters: [])
+      filtered_params = params.reject { |key, _| ['action', 'controller', 'format'].include?(key) }
+      return if (filtered_params.keys & allowed_parameters).length == filtered_params.keys.length
+
+      ## rubocop:disable Metrics/LineLength
+      raise Sushi::Error::UnrecognizedParameterError.new(data: "The given parameter(s) are invalid: #{(filtered_params.keys - allowed_parameters).join(', ')}.")
+      ## rubocop:enable Metrics/LineLength
+    end
+  end
+
+  ##
   # This module accounts for date behavior that needs to be used across all reports.
   #
   # @see #coerce_dates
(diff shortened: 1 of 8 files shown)


app/models/sushi.rb
```diff
@@ -112,6 +112,21 @@ def rescued_last_month_available(*args)
...
 
+  ##
+  # @param params [Hash, ActionController::Parameters]
+  # @param allowed_parameters [Array<String>]
+  #
+  # @return [String]
+  # @raise [Sushi::Error::UnrecognizedParameterError] when any unrecognized params are given.
+  module ParameterValidation
+    def validate_paramaters(params = {}, allowed_parameters: [])
+      filtered_params = params.reject { |key, _| ['action', 'controller', 'format'].include?(key) }
```
jeremyf, line 123:
```ruby
given = params.keys.map(&:to_s) - ['action', 'controller', 'format']
return true if (given & allowed_parameters).length == given.length

 raise Sushi::Error::UnrecognizedParameterError.new(data: "The given parameter(s) are invalid: #{(given - allowed_parameters).join(', ')}.")
```

On the original line 124 you mapped the keys to_s; but if there's a mismatch, you don't remap the keys to string.  This handles that mapping ensuring you're comparing strings to strings..

jeremyf (changes requested):

app/models/sushi.rb
alishaevn: good catch. I hadn't tested for the latter statement and it didn't cross my mind.

alishaevn: also, is there an actual benefit of returning true instead of just returning?

jeremyf: No significant benefit…just a slight "Hey this `validate` method said true" is kind of a nice response (given that Nil is false)

jeremyf (approved):

app/models/sushi.rb
```diff
@@ -112,6 +112,23 @@ def rescued_last_month_available(*args)
     end
   end
 
+  ##
+  # This module provides a common interface for validating parameters.
+  #
+  # @param params [Hash, ActionController::Parameters]
+  # @param allowed_parameters [Array<String>] a list of query parameters AND report filters that are allowed.
```
alishaevn, line 119: @jeremyf asking for your new review due to this change. it also required a change in the specs.

jeremyf: It looks good.  The specs are good to update.

jeremyf (approved):

PR #836, opened 2023-09-30
🐛 Fix PDF.js logic

kirkkwang:
This commit will fix the logic for PDF.js to work when the first file set presenter is not a PDF.  We first select all pdf file set presenters and then select the first one.

### DIFF
--- a/app/helpers/pdf_js_helper.rb
+++ b/app/helpers/pdf_js_helper.rb
@@ -5,6 +5,11 @@ module PdfJsHelper
     "/pdf.js/viewer.html?file=#{path}##{query_param}"
   end
 
+  def pdf_file_set_presenter(file_set_presenters)
+    # currently only supports one pdf per work
+    file_set_presenters.select(&:pdf?).first
+  end
+
   def query_param
     return unless params[:q]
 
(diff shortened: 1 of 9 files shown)


app/helpers/pdf_js_helper.rb
```diff
@@ -5,6 +5,11 @@ def pdf_js_url(path)
     "/pdf.js/viewer.html?file=#{path}##{query_param}"
   end
 
+  def pdf_file_set_presenter(file_set_presenters)
+    # currently only supports one pdf per work
+    file_set_presenters.select(&:pdf?).first
```
ShanaLMoore, line 10: is it ok that this could return nil?

kirkkwang: hmm, it shouldn't return `nil` because it wouldn't should get called if there isn't any pdf file sets

```erb
<% elsif Flipflop.default_pdf_viewer? && presenter.show_pdf_viewer? && presenter.file_set_presenters.any?(&:pdf?) %>
  <%= render 'pdf_js', file_set_presenter: pdf_file_set_presenter(@presenter.file_set_presenters) %>
```

ShanaLMoore: ah, missed that. ok thanks!

ShanaLMoore (approved):

Issue #841, opened 2023-10-06
Refine account settings page to limit options for non-super admins


jillpe:
## Summary
The Account Setting page has too many options for non-super admins, and Pals would like to limit their options. 

<details><summary>Screenshots</summary> 

![image](https://github.com/scientist-softserv/palni-palci/assets/84697174/3517fb7a-8888-4de9-a5e1-bf3c762c2f62)

</details>

Hide certain settings from non-superadmins on the Account settings dashboard page as prescribed by the client.

# Acceptance Criteria

A non super admin user should **NOT** see the following fields: 

- [ ] :~contact_email~ See [comments](https://github.com/scientist-softserv/palni-palci/issues/841#issuecomment-1765434879) section for update
- [ ] :contact_email_to,
- [ ] :analytics_provider
- [ ] :file_acl
- [ ] :s3_bucket,
- [ ] :oai_prefix
- [ ] :oai_sample_identifier,
- [ ] :file_size_limit

# Testing Instructions

Visit the options [here](https://demo.commons-archive.org/admin/account/edit?locale=en), as a super admin and a non super admin. 

super admin login: support@notch8.com => get password from 1Password
non super admin login: user@example.com => testing123

# Expected Behavior Before Changes

<details> <summary> screenshot (non superadmin user) </summary>

![image](https://github.com/scientist-softserv/palni-palci/assets/10081604/d0babdb3-8b6a-4c1a-9b70-c8b8bc25a02e)

</details>

<details> <summary> screenshot (superadmin user) </summary>

![image](https://github.com/scientist-softserv/palni-palci/assets/10081604/1daf1618-9fc9-462f-8c46-4249294ef721)

</details>

# Expected Behavior After Changes
<details> <summary> screenshot (non superadmin user) </summary>

![image](https://github.com/scientist-softserv/palni-palci/assets/10081604/5029eb16-c3db-47a9-8324-ac3fe3a00e31)

</details>

<details> <summary> screenshot (superadmin user) </summary>

![image](https://github.com/scientist-softserv/palni-palci/assets/10081604/d1e3c47a-76fe-4fd3-ba1a-e5527cefb6e8)

</details>

jillpe:
What do you want non-super admins to see?

ndroark:
Partially, we're not sure what some of these fields are for, and thus we're not sure what could potentially allow users to break systems by changing settings. 

The ones we don't know about are:

- Analytics Provider
- File acl
- OAI prefix
[...]

PR #848, opened 2023-10-10
🎁 Limit settings to superadmins

kirkkwang:
This commit will hide certain settings from non-superadmins on the Account settings dashboard page as prescribed by the client.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/841

# Story

Refs #issuenumber

# Screenshots / Video

<details>
<summary>As superadmin</summary>

image

</details>

<details>
<summary>As non-superadmin</summary>

image

</details>

### DIFF
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -112,7 +112,7 @@ module AccountSettings
 
   def public_settings(is_superadmin: false)
     settings = all_settings
-    settings = superadmin_settings(is_superadmin) unless is_superadmin
+    settings = superadmin_settings unless is_superadmin
     settings.reject { |k, v| Account.private_settings.include?(k.to_s) || v[:disabled] }
   end
 
@@ -238,7 +238,7 @@ module AccountSettings
       # rubocop:enable Style/RedundantSelf
     end
 
-    def superadmin_settings(is_superadmin)
+    def superadmin_settings
       all_settings.reject { |k, _v| Account.superadmin_settings.include?(k.to_sym) }
     end
 end
--- a/app/models/concerns/account_settings.rb
+++ b/app/models/concerns/account_settings.rb
@@ -110,8 +110,10 @@ module AccountSettings
   end
   # rubocop:enable Metrics/BlockLength
 
-  def public_settings
-    all_settings.reject { |k, v| Account.private_settings.include?(k.to_s) || v[:disabled] }
+  def public_settings(is_superadmin: false)
+    settings = all_settings
+    settings = superadmin_settings(is_superadmin) unless is_superadmin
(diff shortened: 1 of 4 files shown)


app/models/concerns/account_settings.rb
```diff
@@ -110,8 +110,10 @@ def solr_collection_options
   end
   # rubocop:enable Metrics/BlockLength
 
-  def public_settings
-    all_settings.reject { |k, v| Account.private_settings.include?(k.to_s) || v[:disabled] }
+  def public_settings(is_superadmin: false)
+    settings = all_settings
+    settings = settings.reject { |k, _v| Account.superadmin_settings.include?(k.to_sym) } unless is_superadmin
```
ShanaLMoore, line 115: only a suggestion, we could extract this out into its own method.

ShanaLMoore (approved):

app/models/concerns/account_settings.rb
```diff
@@ -110,8 +110,10 @@ def solr_collection_options
   end
   # rubocop:enable Metrics/BlockLength
 
-  def public_settings
-    all_settings.reject { |k, v| Account.private_settings.include?(k.to_s) || v[:disabled] }
+  def public_settings(is_superadmin: false)
```
ShanaLMoore, line 113:
```suggestion
def public_settings(is_superadmin: false)
  settings = is_superadmin ? superadmin_settings(is_superadmin) : all_settings
  filter_out_private_and_disabled_settings(settings)
end

private

def filter_out_private_and_disabled_settings(settings)
  settings.reject do |key, value|
    Account.private_settings.include?(key.to_s) || value[:disabled]
  end
end

```

ShanaLMoore: ^^ the above isn't quite right but something along those lines is what I was thinking

app/models/concerns/account_settings.rb
```diff
@@ -110,8 +110,10 @@ def solr_collection_options
...
+    settings = all_settings
+    settings = superadmin_settings(is_superadmin) unless is_superadmin
```
ShanaLMoore, line 115:
```suggestion
    settings = superadmin_settings unless is_superadmin
```

Sorry I hadn't refreshed this page and see you applied similar changes already. I don't think you will need to pass is_superadmin to superadmin_settings though. Rubocop is complaining that it's an unused variable.

app/models/concerns/account_settings.rb
```diff
@@ -235,4 +237,8 @@ def reload_analytics
       Hyrax.config.analytics = Hyrax::Analytics.config.valid?
       # rubocop:enable Style/RedundantSelf
     end
+
+    def superadmin_settings(is_superadmin)
```
ShanaLMoore, line 241:
```suggestion
    def superadmin_settings
```

Issue #853, opened 2023-10-12
Hack to load video


ShanaLMoore:
# Summary

Nic has a client demo and noticed that the AV was not loading his video, the day before the meeting. He had assumed the 59 minute video had finished processing because he uploaded it the day before.

After looking into it, Kirk and Shana discovered that the video was still processing.

Implement a hack in the meantime, to help Nic be successful with his demo.

TODO: The iiif manifest should reference the downloadable link instead of the video derivative. 

related: 
- https://github.com/scientist-softserv/palni-palci/issues/852

# Screenshot

![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/accd11a4-31af-40f7-b1fb-3dd4080e8f11)

![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/49f6d37f-c376-4d53-b93d-656579aa283f)

# Notes

related convo: 
https://assaydepot.slack.com/archives/C0313NKC08L/p1697141000008749

This workaround falls apart if the uploaded video is a raw video file where the size could be enormous. That is not ideal for streaming. Nic confirmed that raw videos/streaming is not encouraged.

ShanaLMoore:
![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/64e22cfd-c104-4272-bd88-50e7e68c5f75)

PR #855, opened 2023-10-12
HOT FIX FOR LONG VIDEO PROCESSING TIMES DO NOT MERGE

kirkkwang:
This commit will set the IIIF manifest URI for videos to be the Hyrax download video.  This is a temporary fix until we can get the video processing speeds faster.  It won't make a difference if the video that was ingested was already processed, however, this workaround falls apart if the uploaded video is a raw video file where the size could be enormous.  That is not ideal for streaming.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/853

### DIFF
--- a/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb
+++ b/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb
@@ -42,11 +42,18 @@ module Hyrax
           height = solr_document.height&.try(:to_i) || 240
           duration = conformed_duration_in_seconds
           IIIFManifest::V3::DisplayContent.new(
-            Hyrax::IiifAv::Engine.routes.url_helpers.iiif_av_content_url(
-              solr_document.id,
-              label: label,
-              host: request.base_url
-            ),
+            # Hyrax::IiifAv::Engine.routes.url_helpers.iiif_av_content_url(
+            #   solr_document.id,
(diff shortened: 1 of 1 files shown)


ShanaLMoore (approved):

ShanaLMoore:
cc @kirkkwang Tell me more about the "do not merge" warning (mostly because I don't recall what we all agreed on).

I thought the hot fix deployed would stay live until Rob fixes the processing speeds. Since we are deploying next Thursday I thought it'd be ok to merge this into main, because we're unsure how long it'll take Rob to address this issue. I left a note in Rob's ticket about reverting this work upon his completion.

kirkkwang:
This one is just a hot fix branch that is made off of a production SHA, there's another PR that has the code, I can't remember if it's merged already or not.  Thinking back that wasn't really needed, but this branch can be thrown away.

kirkkwang:
This one is just a hot fix branch that is made off of a production SHA, there's another PR that has the code, I can't remember if it's merged already or not.  Thinking back that wasn't really needed, but this branch can be thrown away.

ShanaLMoore:
gotcha. I confirmed that code is already in [main](https://github.com/scientist-softserv/palni-palci/blob/main/app/presenters/concerns/hyrax/iiif_av/displays_content_decorator.rb#L56)

Issue #748, opened 2023-08-22
Ability to sort works and collections on dashboard


jillpe:
## Summary
Pals would like the ability to sort works and collections on the dashboard. They would like to be able to sort both by:
- date uploaded
- date modified
- date created
- depositor (a-z)
- creator (a-z)

<details><summary>Screenshots</summary>

![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/5b3265e4-987a-41df-bc6c-98aeee366d29)

![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/1a50f565-a4cd-45a1-8fcd-3a8a17fbb7eb)

</details>

## Accepted Criteria

- [ ] As a user, I can sort works on the 'Works' dashboard page by ~date uploaded~, `date modified`, `date created`, `depositor (a-z)`, and `creator (a-z)`
- [ ] As a user, I can sort collections on the 'Collections' dashboard page by `date uploaded`, `date modified`, `date created`, `depositor (a-z)`, and `creator (a-z)`

jillpe:
Do you have a preference on where the sort button should go? Or do you want a developer to decide?

ndroark:
I think the space between Show # per page and Search makes sense, if that's an option

ShanaLMoore:
# QA RESULTS: 

Tested on: https://dev.commons-archive.org/dashboard/my/collections?locale=en

## COLLECTIONS 
- [X] As a user, I can sort collections on the 'Collections' dashboard page by date uploaded, date modified, date created, depositor (a-z), and creator (a-z)



Demo: https://share.zight.com/o0uKe9nO




![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/eec25fe2-0359-4602-8978-0a893fd988a4)





## WORKS
- [X] As a user, I can sort works on the 'Works' dashboard page by date uploaded, date modified, date created, depositor (a-z), and creator (a-z)


DEMO: https://share.zight.com/Wnupx9Zl




![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/2ee3d7b5-d841-4558-a999-ca0b70bfaefb)

ShanaLMoore:
Why is date uploaded missing from the collection dashboard? 

Per @kirkkwang - 

``` ruby
        # Collections don't seem to have a date_uploaded_dtsi nor date_modified_dtsi
        # we can at least use the system_modified_dtsi instead of date_modified_dtsi
[...]

PR #867, opened 2023-10-17
I748 sort works and collections on dashboard

kirkkwang:
# Story

[🎁 Add sort on dashboard works](https://github.com/scientist-softserv/palni-palci/commit/3afa0dd01ed4f3acf252da95cd48e7aed070c2d8) 

This commit will add the ability to sort on the dashboard works pages,
to include dashboard works and my works.

[🎁 Add ability to sort dashboard collections](https://github.com/scientist-softserv/palni-palci/commit/03efb9038a43a694fc8f8e06c96e842b4a5ec8f7) 

This commit will add the ability to sort collections on the dashboard.
The collections work differently than the works so we have to make sure
the sort field is added to the search builder.  Because when you click
into a collection can view the works attached to the collection, we have
to approach it differently because otherwise the sort fields from the
dashboard index page will also be applied to the show page.  We decorate
Blacklight helpers to check the controller action name as our
conditional.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/748

# Screenshots / Video

https://github.com/scientist-softserv/palni-palci/assets/19597776/36414e44-72a2-4617-b7ae-c28354efb587

# Note
Reindexing is needed to get sorting by depositor and creator working.

### DIFF
--- a/app/helpers/blacklight/configuration_helper_behavior_decorator.rb
+++ b/app/helpers/blacklight/configuration_helper_behavior_decorator.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+# OVERRIDE Blacklight v6.25.0 to add custom sort fields while in the dashboard for both collections and works
+
+module Blacklight::ConfigurationHelperBehavior
+  def active_sort_fields
+    if dashboard_controller_and_index_action?
+      dashboard_sort_fields.sort_fields.select { |_sort_key, field_config| should_render_field?(field_config) }
+    else
+      blacklight_config.sort_fields.select { |_sort_key, field_config| should_render_field?(field_config) }
+    end
+  end
+
+  def dashboard_controller_and_index_action?
+    (params[:controller].include?('hyrax/my/collections') || params[:controller].include?('hyrax/dashboard/collections')) && action_name == 'index'
+  end
+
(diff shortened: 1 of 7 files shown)


app/helpers/blacklight/configuration_helper_behavior_decorator.rb
```diff
@@ -0,0 +1,52 @@
...
+                   end
+    field_config ||= Blacklight::Configuration::NullField.new(key: key)
+
+    field_label(
+      :"blacklight.search.fields.sort.#{key}",
+      (field_config&.label),
+      key.to_s.humanize
+    )
+  end
+end
```
jeremyf, line 52: I'm missing when we inject this decoration.  Also, this is something that would definitely be good to include a test.

kirkkwang: @jeremyf How about the name `configuration_helper_behavior_overrides.rb`?  I can't name it `configuration_helper_behavior.rb` because it doesn't have all the methods of the original helper from blacklight.

ShanaLMoore (approved):
After exhausting many options, and in consideration of our deadline, we are saving adding additional specs for another time.

PR #869, opened 2023-10-17
Trying to get solr image to work on amd64 and therefor lint action to work again.

orangewolf:


ShanaLMoore:
thanks for fixing this, I was trying to figure this out yesterday and definitely wasn't on the right path 🙃

PR #893, opened 2023-10-31
 🐛 Fix misspelling on depositor

laritakr:
# Story

Fix misspelling of depositor so sort works

Refs
- https://github.com/scientist-softserv/palni-palci/issues/748

# Screenshots / Video

<details>
<summary></summary>

</details>

# Notes
Requires a reindex of both works and collections


ShanaLMoore (approved):

ShanaLMoore:
Do we still want to normalize the creator value, since we'll have to reindex anyways @laritakr ?

laritakr:
> Do we still want to normalize the creator value, since we'll have to reindex anyways @laritakr ?

I don't know if it matters? The majority should be all lower case... very few people input email addresses with mixed case. I guess you can ask Nic, but I would probably not.

PR #896, opened 2023-11-01
Update reindex_collections_job.rb

ShanaLMoore:
# Notes

Is the reindex rake task creating slow jobs because it's missing this line? It's taking hours to complete a single job.

`collection.try(:reindex_extent=, Hyrax::Adapters::NestingIndexAdapter::LIMITED_REINDEX)`

as seen here: 

https://github.com/scientist-softserv/palni-palci/blob/main/app/jobs/reindex_collections_job.rb#L10

Screenshot 2023-11-01 at 11 14 25 AM

ref: https://github.com/scientist-softserv/palni-palci/issues/901

### DIFF
--- a/app/jobs/reindex_collections_job.rb
+++ b/app/jobs/reindex_collections_job.rb
@@ -3,13 +3,16 @@
 class ReindexCollectionsJob < ApplicationJob
   def perform(collection_id = nil)
     if collection_id.present?
-      collection = Collection.find(collection_id)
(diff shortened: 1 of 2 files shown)


kirkkwang (approved):

ShanaLMoore:
Converting to draft. If we fix the source of the problem (making PALS use the graph indexer), this fix shouldn't be needed.

PR #899, opened 2023-11-01
:gift: Add HYRAX_USR_SOLR_GRAPH_NESTING env variable

ShanaLMoore:
Since PALS is on 3.6 version of hyrax it has graph_indexer to improve the speed of queries and reindexing. However PALS does not have the env var set to enable it in the various environments. There was an ENV['HYRAX_USE_SOLR_GRAPH_SETTING'] being set in hyrax.rb, however it doesn't look like it was being honored. See screenshot of error message from production.

Screenshot 2023-11-01 at 11 14 25 AM


kirkkwang (approved):

laritakr (changes requested):
~This change would break the functionality I fixed in https://github.com/scientist-softserv/palni-palci/pull/902 as the query for the all_collections route relies on the nested indexing. I don't think this is a change we want to implement. Verifying that we are using the graph indexer is a much better solution.~ Put this comment on the wrong ticket!

ShanaLMoore:
@laritakr merging is blocked without yoru re - review. But to be clear I'll be merging this and #902 to test on staging

laritakr (approved):

PR #900, opened 2023-11-01
Add limited reindex to reindex works job

ShanaLMoore:
related PR:  https://github.com/scientist-softserv/palni-palci/pull/896

This PR includes changes to ensure the reindex job is using limited reindex. Otherwise, this job could potentially take forever to run. 

ref: https://github.com/scientist-softserv/palni-palci/issues/901


ShanaLMoore:
Converting to draft. If we fix the source of the problem (making PALS use the graph indexer), this fix shouldn't be needed.

Issue #901, opened 2023-11-01
🐛 Reindexing collections takes forever


ShanaLMoore:
In our latest hotfix to pals, reindexing Collections was required. Running the rake task in staging worked fine, but a single job would take hours to complete in prod (maybe because there's not as much data as prod). 

I stopped and deleted all of those jobs in prod. The rest of the enqueued jobs finished rapidly so the slowness seems to be related to the reindex job; there's probably some nested indexing issue going on. 

## NOTES 

Relevant slack convo: https://assaydepot.slack.com/archives/C0311DNF3MG/p1698810560689329

ShanaLMoore:
Thanks @laritakr your changes helped tremendously here. We were able to reindex works and collections overnight.

PR #902, opened 2023-11-01
🐛 Fix top level collections

laritakr:
# Story

Fixing the app to use the graph indexer means that the query for top level collections will not work. It relied on an index field, `nesting_collection__deepest_nested_depth_isi` which was added by the nesting indexer. With the use of the graph indexer, this term no longer exists. 

This is a minor issue, as it ONLY affects the `all_collections` route for the homepage when using the `institutional_repository` theme, and this route is not linked anywhere. Without this change, the list of collections on this path would have been empty once collections are reindexed, as the queried term would no longer exist.

A simpler solution is to identify a top level collection as one which is not a member of any other collection. Along with this change, come code duplication is also being cleaned up.

Refs
- https://github.com/scientist-softserv/palni-palci/issues/901

Once a collection is reindexed, it would no longer be identified as a top level collection.

Collections at the `all_collections` route will include all collections which do not belong to another collection, and will no longer rely on the obsolete nesting indexer term to determine this inclusion.

# Screenshots / Video

<details>
<summary></summary>

Collection appears in all_documents view
Screenshot 2023-11-01 at 4 11 59 PM

## Solr document does not include the nesting indexer terms
Screenshot 2023-11-01 at 4 12 31 PM

</details>

# Notes

### DIFF
--- a/app/controllers/hyrax/homepage_controller.rb
+++ b/app/controllers/hyrax/homepage_controller.rb
@@ -90,7 +90,7 @@ module Hyrax
       end
 
       def load_top_level_collections(colls)
-        colls.select{ |c| c['member_of_collection_ids_ssim'] == nil }
+        colls.select{ |c| c['member_of_collection_ids_ssim'].nil? }
       end
 
       # Return 6 collections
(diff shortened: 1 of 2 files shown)


ShanaLMoore (approved):
Thanks for noticing and taking care of this!

PR #925, opened 2023-12-06
🐛 HACK: Disable mp4 and webm derivative generation

jeremyf:
With this commit we are not generating time-based derivatives for
movies; instead generating only the thumbnail.

The reason is that as of <2023-12-06 Wed> the mp4 generation is
extremely slow (and perhaps the webm; hence the 🔨 for this 🐛).

Related to:

- https://github.com/scientist-softserv/palni-palci/issues/924


ShanaLMoore:
Thanks Jeremy. That was much easier than I thought! 

I live edited the production server with this fix in hopes of immediate changes.

Issue #911, opened 2023-11-30
Embed codes for media (see Atla instance)


jillpe:
## Summary
relevant tickets
- https://github.com/scientist-softserv/atla-hyku/issues/9

Pals would like to have the ability to embed YouTube videos onto a work show page, and they would like it to be implemented like Atla. 

The tagged PRs are:

- https://github.com/scientist-softserv/atla-hyku/pull/34
- https://github.com/scientist-softserv/atla-hyku/pull/46

## Task Breakdown

- [x] - [Implement video embed for Generic Works](https://github.com/scientist-softserv/palni-palci/pull/922)
- [x] - Implement video embed for other work types
  - [x] [CDL](https://github.com/scientist-softserv/palni-palci/pull/927)
  - [x] [Image](https://github.com/scientist-softserv/palni-palci/pull/928/files#diff-cb716bffdfff686cdbc6e5a7ebec30e8638f516a6aa4b512c381447570d7ba3c)
  - [x] [OER](https://github.com/scientist-softserv/palni-palci/pull/931) 
  - [x] [ETD](https://github.com/scientist-softserv/palni-palci/pull/929) 
- [x] - [Run translations](https://github.com/scientist-softserv/palni-palci/pull/926)
- [x] - [Add embedded video to other theme's show pages](https://github.com/scientist-softserv/palni-palci/pull/930)
- [x] when a user submits an invalid url, an error message should get thrown
  - [x] [The error message should support the various locales 
](https://github.com/scientist-softserv/palni-palci/pull/931)
## Acceptance Criteria
- [ ] As a user, I can add a link to a youtube video I would like to embed on the edit work form
- [ ] - [ ] As a user, I can add a link to a vimeo video I would like to embed on the edit work form
- [ ] As a user, i can add a column to bulkrax csv import that allows me to import a embedded video url (youtube and/or vimeo) that will be used in an iframe.
- [ ] As a user, I can visit that work's work-show page to see & play the embedded videos iframe
- [ ] Client is sent the correct info for what column to add to bulkrax csvs.
- [ ] Atla PRs that are related to this are added to Pals
- [ ] all work types are able to embed videos
- [ ] all themes show the embedded video
- [ ] embed links must be valid. A user cannot submit the form if not.

### BULKRAX

Import a sample CSV, but add a column called video_embed. The values will be the embed links from youtube and/or vimeo. 
After the jobs finish processing, navigate to the works' show pages to confirm the videos are there and playable. 

### OTHER

[...]

PR #931, opened 2023-12-06
🎁 Add video embed to OER Work Type

ShanaLMoore:
This PR manually adds video embed to oer model, form files, and show partial.

It also includes translations. 

Issue:
- https://github.com/scientist-softserv/palni-palci/issues/911

# OER show page (OER does not use the base/show.html.erb like the other model types)

Vimeo example: https://player.vimeo.com/video/467264493?h=b089de0eab

image


ShanaLMoore:
UPDATE: 

I'm noticing that editing the other work types is presenting the same behavior. For the sake of time I may undo the module work and proceed by putting the video embed code directly onto the models and form files, like done in this PR.

kirkkwang (approved):

PR #959, opened 2024-01-11
Fixing reshare some more

orangewolf:
# Story

There are issues with reshare. If a url is opened from an email or directly pasted in to the URL bar, it works fine, but if it is from a clicked link on a web page it will not work. 

Going to http://old.notch8.com/test.html and clicking either test link will take you through Reshare login (see 1password) and then to an error page on the reshare side.

Going to http://old.notch8.com/test.html and clicking either test link will take you through Reshare login (see 1password) and then to our side where the work loads.


orangewolf:
@jeremyf I ended up moving to cookies because the session kept getting dropped. while I was at it, I simplified things to be more devise/omniauth typical. This is going up to prod because the reshare folks are getting frustrated but still needs review and merge and all that jazz

jeremyf (approved):

app/controllers/application_controller.rb
```diff
@@ -65,37 +102,18 @@ def is_staging
       ['staging'].include?(Rails.env)
     end
 
-    ##
-    # Extra authentication for palni-palci during development phase
-    def authenticate_if_needed
```
jeremyf, line 70: There remains a `before_action :authenticate_if_needed` declaration on line 21.  Yet we're removing this method.  Which should it be?  My understanding is that without this in place, the staging environment will no longer require HTTP Auth.

jeremyf (changes requested):

jeremyf (approved):

PR #961, opened 2024-01-17
🎁 Add loading splash screen and speed up redirect

kirkkwang:
The clients requested a loading screen that covers the entire page so users can't click on anything while the redirect is happening.

https://github.com/scientist-softserv/palni-palci/assets/19597776/9efc588f-eaff-41f1-a3c6-0380da8a9514

### DIFF
--- a/app/jobs/create_group_and_add_members_job.rb
+++ b/app/jobs/create_group_and_add_members_job.rb
@@ -22,7 +22,7 @@ class CreateGroupAndAddMembersJob < ApplicationJob
     child_works_count = work.members.select { |member| member.is_a?(child_model) }.count
 
     if page_count == child_works_count
-      group = Hyrax::Group.create(name: work.id)
+      group = Hyrax::Group.find_or_create_by!(name: work.id)
       work.read_groups = [group.name]
 
       work.members.each do |member|
--- a/app/jobs/create_group_and_add_members_job.rb
+++ b/app/jobs/create_group_and_add_members_job.rb
@@ -7,6 +7,8 @@ class CreateGroupAndAddMembersJob < ApplicationJob
 
   def perform(cdl_id, retries = 0)
     work = Cdl.where(id: cdl_id).first
+    return if work.nil?
+
     page_count = work.file_sets.first.page_count.first.to_i
     child_model = work.iiif_print_config.pdf_split_child_model
     child_works_count = work.members.select { |member| member.is_a?(child_model) }.count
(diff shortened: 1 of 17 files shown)


app/models/cdl.rb
```diff
@@ -50,4 +55,24 @@ class Cdl < ActiveFedora::Base
   # including `include ::Hyrax::BasicMetadata`. All properties must
   # be declared before their values can be ordered.
   include OrderMetadataValues
+
+  private
+
+    def create_group_and_add_members
+      return if is_child
+
+      # IiifPrint::Jobs::CreateRelationshipsJob has a 10 min timeout
```
jeremyf, line 64: Wondering about the timing of this?  Should we instead key off of the `CreateRelationshipsJob`?  We'll need to talk this over.

jeremyf (changes requested):

app/jobs/create_group_and_add_members_job.rb
```diff
@@ -0,0 +1,51 @@
...
+  def perform(cdl_id, retries = 0)
+    work = Cdl.where(id: cdl_id).first
+    return if work.nil?
+
+    page_count = work.file_sets.first.page_count.first.to_i
+    child_model = work.iiif_print_config.pdf_split_child_model
+    child_works_count = work.members.select { |member| member.is_a?(child_model) }.count
+
+    if page_count == child_works_count
+      group = Hyrax::Group.create(name: work.id)
```
jeremyf, line 25: Given that we have retry attempts favoring a find or create will allow us to:

1. Avoid name collisions
2. Allow for greater fault tolerance.

```suggestion
      group = Hyrax::Group.find_or_create_by!(name: work.id)
```

app/models/cdl.rb
jeremyf: I think we should decorate the `IiifPrint::Jobs::CreateRelationshipsJob` to conditionally launch this job.

kirkkwang: but the job itself has a check built in as well as retries if conditions aren't met, but I do agree with keying off of `CreateRelationshipsJob`.

app/jobs/create_group_and_add_members_job.rb
```diff
@@ -0,0 +1,50 @@
...
+        assign_read_groups(member, group.name)
+      end
+
+      work.save
+      group.save
+    else
+      return if retries > RETRY_MAX
+
+      retries += 1
+      CreateGroupAndAddMembersJob.set(wait: 10.minutes).perform_later(cdl_id, retries)
```
jeremyf, line 37: Yes this one gets to retry!

jeremyf (approved):

Issue #948, opened 2023-12-13
Bug: Banner images in search tenant affecting other tenants


ndroark:
I'm experiencing an issue where updating a banner image in our search tenant ([search.hykucommons.org](http://search.hykucommons.org/)) changes the banner image in a different tenant ([uindy.hykucommons.org](http://uindy.hykucommons.org/)) and vice-versa.

This leaves us unable to set banner images in two public-facing tenants.

ndroark:
We should add something like "Banner images in tenants cannot be overriden by settings in any other tenant"

jillpe:
SoftServ QA: ✅ 

<details><summary>https://dev.commons-archive.org/?locale=en</summary>


![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/d9c6bd07-f48c-4604-ada4-e9cee3ea40ec)


</details>

<details><summary>https://bloomsburg.commons-archive.org/?locale=en</summary>


![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/376d1cdb-97ad-44fb-9332-9c40b2e9bc3a)


</details>

ndroark:
Not sure if it's within the scope of this ticket, but the alt text doesn't seem to be required now?

kirkkwang:
@ndroark ready for another look

Issue #746, opened 2023-08-22
Enhance banner image customization


jillpe:
## Summary
Pals would like to be able to preview the banner customization before saving, as well as be able to select the positioning and cropping of the image **after** it is uploaded.

## Accepted Criteria

- [ ] As a user, I can preview my banner customization before saving it
- [ ] As a user, after an image is uploaded for the banner image customization, I can adjust the positioning and cropping

## Testing Instructions

- From the admin dashboard, navigate to the [banner image tab](https://demo.commons-archive.org/admin/appearance?locale=en#banner_image) in the appearance settings. 
- Import a banner image. Sample - 

![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/0931c738-694d-45ea-922e-7d20c83892d1)

- [ ] a user should be able to crop the image before saving

- Save the banner image and visit the home page.
- The home page's banner should update to the imported/cropped banner photo 
- 
**- NOTE: The home page's banner is dynamic/responsive so it may not portray exact dimensions as cropped. Nic is ok with this for now as it still reduces the efforts for the users.** 

- Navigate back to the banner image settings. A user should be able to click crop to make adjustments to the previously cropped image. **NOTE: if a user wants to recrop from the original image they will have to reupload - Nic is OK w this for now.** 
- Check the homepage banner for changes

- Navigate back to the banner image settings. Click the remove button. 
- [ ] the banner image should go back to the default. Confirm your uploaded image is no longer seen on the homepage banner.

## Screenshots

![Image](https://github.com/scientist-softserv/palni-palci/assets/10081604/834787a0-8e66-4db3-8fd8-9d640588a80d)

ShanaLMoore:
# DEV NOTES/Resources

https://stackoverflow.com/questions/21840899/preview-image-before-upload-with-carrierwave

ShanaLMoore:
# QA Results: ❌ 

Tested on: https://demo.commons-archive.org/?locale=en

Initially the homepage updated to what I selected. However, when I re upload/crop and click save, the homepage banner doesn't appear to be updating. demo: https://share.zight.com/rRumGZYe

Not a cause for kickback but I'd argue crop button should not be displayed on the initial upload (since it's activated automatically).

ShanaLMoore:
I see the issue. silly mistake. Closing this PR

ShanaLMoore:
# QA RESULTS: Pass ✅ ⚠️ 

[...]

PR #981, opened 2024-03-04
🐛 Fix cropped banner images being overwritten

kirkkwang:
When cropper.js intercepts an updated banner image file and crops it, the original filename gets changed to cropped.jpg which gets saved to disk.  When another tenant crops their banner image, it also gets saved as cropped.jpg and overwrites the previous tenant's banner image.  In this commit, we prepend the tenant name so it would look something like `shared-cropped.jpg` and `uindy-cropped.jpg` instead and avoids the files from being overwritten.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/948
  - https://github.com/scientist-softserv/palni-palci/issues/746

Banner images were being overwritten.

Banner images have unique names based off the tenant so overwriting the name shouldn't be possible.

# Screenshots / Video

## Before
UIndy tenant has the banner image with filename being `cropped.jpg`
image

Search tenant also has the same filename of `cropped.jpg`
image

## After
The tenant is prepended to the filename
image

Another example
image

### DIFF
--- a/app/controllers/users/omniauth_callbacks_controller.rb
+++ b/app/controllers/users/omniauth_callbacks_controller.rb
@@ -30,7 +30,7 @@ module Users
         # where we want a JS-based redirect to go.
         render 'complete', locals: { redirect_to_url: url || hyrax.dashboard_path }
       else
-        redirect_to root_path, flash: {error: 'Not able to log in user. #{@user.errors.full_messages}'}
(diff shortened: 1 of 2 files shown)


ShanaLMoore (approved):
Nice catch. We'll definitely want to contribute this fix back.

Issue #783, opened 2023-09-12
Issue with adding collections as subcollections


ndroark:
## Summary
I am trying to add collections to existing collections (while logged in as a superadmin), but after I select the subcollection from the dropdown list, I get this error message. Occuring in franklin.hykucommons.org.

![Image](https://user-images.githubusercontent.com/117871943/267379252-095de82d-90d5-43e2-885e-cd374b7e3036.png)

## Acceptance Criteria
- [ ] Any user with collection manager level permissions can add collections to existing collections through the dashboard
- [ ] This code has been contributed to the double combo or main branch (the longest running branch). Note that this rake task doesn't work as is on hyku since hyrax is v5+. 

# NOTE

This ticket was unrelated to permissions. If collection manager permissions continue, this should be a new ticket.

# Testing

Verify that the 3 affected collections in Franklin tenant now show the full list of valid collections to be added as subcollections. The 3 affected collections were [College Presidents](https://franklin.hykucommons.org/dashboard/collections/3ced02ba-d2dd-49fe-af48-42207f013ce4?locale=en), [Franklin College in the News](https://franklin.hykucommons.org/dashboard/collections/0c3005c0-f126-4b7a-ac77-20a2b725f75b?locale=en), and [Lasting Legacies](https://franklin.hykucommons.org/dashboard/collections/39e102cc-551c-4d41-8f26-2d87190ac39f?locale=en).

To test in staging, you would have to find an old collection (prior to 2024), create a new collection, and verify that subcollections can be added to each other.

bkiahstroud:
@ndroark can you confirm whether or not this is still an issue? 

Based on the screenshot, the intended parent collection is `Lasting Legacies: Franklin College 'Spirit of Community' Collection`, and the intended subcollection is `Judi Warren '80`. 

Looking at the [parent collection now](https://franklin.hykucommons.org/dashboard/collections/39e102cc-551c-4d41-8f26-2d87190ac39f?locale=en), `Judi Warren '80` is a subcollection:  

<details><summary>Screenshot 1</summary>

<img width="1101" alt="image" src="https://github.com/scientist-softserv/palni-palci/assets/32469930/614d178c-ba63-419a-b4ba-ddcf0bfa5d38">

</details>

Though I can tell from the URL that the Collection ID is different, maybe it was recreated? 

I created two test collections in the franklin tenant and was able to add one as a subcollection to the other: 

<details><summary>Screenshot 2</summary>

[...]

PR #982, opened 2024-03-06
Rake task for updating collection_type_gid

laritakr:
# Story

Refs https://github.com/scientist-softserv/palni-palci/issues/783

The structure of the stored collection type gid changed in Hyrax 3.0

A rake task is needed to update the collection_type_gid that is stored on the collection as part of the upgrade from Hyrax 2.x process. This references the logic from the rake task that Hyrax has in versons 3 & 4, and updates it to work within Hyku tenants.

The definition of collection_type_gid= changed in Hyrax 5.0. This rake task is known to work for updates to versions prior to Hyrax 5. Functionality for later versions is unknown.

Collections created in Hyrax 2 cannot be added as parent collections or subcollections of collections created in Hyrax 3 or later.

After running the rake task, collections created in Hyrax 2 have collection type gids that are compatible with the Hyrax 3 collections.

# Screenshots / Video

<details>
<summary></summary>

Screenshot 2024-03-06 at 3 21 04 PM

</details>

# Notes

The collections in production have been updated. Collections in staging have not been updated.


lib/tasks/collection_type_global_id.rake
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+namespace :hyrax do
+  namespace :collections do
+    desc 'Update CollectionType global id references for Hyrax 3.0.0'
+    # Note: the definition of collection_type_gid= changed in Hyrax 5.0. This
+    # rake task is known to work for updates to versions prior to Hyrax 5.
+    # Use for later versions is unknown. 
+    task update_collection_type_global_ids: :environment do
+      # Account.find_each do |account|
```
bkiahstroud, line 9: Looks like this was commented out by mistake.

```suggestion
      Account.find_each do |account|
```

lib/tasks/collection_type_global_id.rake
```diff
@@ -0,0 +1,31 @@
...
+        next if account.search_only.eql? true
+        next if account.is_public.eql? false
```
bkiahstroud, line 11: Why are we ignoring private tenants?

laritakr: I will remove that. I added that code when I verified that all collections got updated in the public tenants, because running the code the first time crashed my rails console. I wanted to run a smaller subset the second time over, just to ensure that it looked like everything was corrected. But you're right that in general we'd want to run for all tenants.

jeremyf (approved):

Issue #436, opened 2023-03-22
Bug: Resource type requirement breaking bulkrax imports on collections and ETDs


crisr15:
# Summary

On #404  we made Resource Type a required field on all models, instead of just worktypes.  And we did not catch that on ETD the field name for Resource Type is "Learning Resource Type" (Just Type as the label in the UI). And that it has a slightly different vocabulary.  

When I add a column for resource type to the importer, it works, and if the resource type and learning resource type are different controlled vocabulary terms, it maps both to Learning Resource Type (since Resource Type isn't a field in the ETD worktype).
edit. And if they are the same, it lists the term twice in Learning Resource Type.  

Sample import: https://delete-me-when-done.hykucommons.org/importers/1?locale=en

Slack message for additional context

Ref:

- https://github.com/scientist-softserv/palni-palci/issues/404

# Analysis

Bulkrax has an implicit assumption that the mapping fields are generally available; but we only map onto the Factory class (e.g. Work, FileSet, and Collection) the fields that exist.

Below is an excerpt from the `lib/configuration/bulkrax.rb`.  The ETD model has a `resource_type` field.  However the OER model has `learning_resource_type`.  When we’re importing rows, if the row were to be for an OER and had a value for `resource_type`, we’d skip adding that property.

```ruby
'learning_resource_type' => { from: ['learning_resource_type'], split: '\|' },
'resource_type' => { from: ['type'], split: '\|' },
```

Fundamentally, before we begin fully processing a row, we need to know what it’s underlying Factory class is; then we can perform data validations by looking at the intersection of:

-   `importerexporter.mapping.keys`
-   `Bulkrax.required_elements`

See [Bulkrax::ApplicationParser#required\_elements](https://github.com/samvera-labs/bulkrax/blob/e1d8a4e7f80074908991907b7aeaa38ba865258d/app/parsers/bulkrax/application_parser.rb#L279-L291); this is on the application parser, but missing is the consideration for the Factory class.

```ruby
def required_elements
  matched_elements = ((importerexporter.mapping.keys || []) & (Bulkrax.required_elements || []))
  unless matched_elements.count == Bulkrax.required_elements.count
    missing_elements = Bulkrax.required_elements - matched_elements
    error_alert = "Missing mapping for at least one required element, missing mappings are: #{missing_elements.join(', ')}"
    raise StandardError, error_alert
  end
  if Bulkrax.fill_in_blank_source_identifiers
    Bulkrax.required_elements
  else
[...]

PR #984, opened 2024-03-07
🎁 Add a more nauced approach to required elements

kirkkwang:
# Story

Prior to this commit we were blanket checking CSV recores for the presence of title and resource_type.  The problem with this approach was that all models were being checked for these fields while not all models required these fields, such as the Collection and FileSet models.

In this commit we leverage the exisiting `Bulkrax::CsvParserDecorator` to check each model's actual required fields according to their respective forms (minus the FileSet which do not have required fields) and add logic to check any custom parser mappings according to the bulkrax initializer and also any numbered fields.  The error message was also changed to indicate which model was missing what required field(s).

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/436

# Screenshots / Video

<details>
<summary>Oer</summary>

image

</details>

<details>
<summary>Collection</summary>

image

</details>

<details>
<summary>FileSet</summary>

image

</details>

### DIFF
--- a/app/parsers/bulkrax/csv_parser_decorator.rb
+++ b/app/parsers/bulkrax/csv_parser_decorator.rb
@@ -1,41 +1,66 @@
 # frozen_string_literal: true
 
-# OVERRIDE BULKRAX 3.1.2 to include oer specific methods
+# OVERRIDE BULKRAX 5.3.0 to include oer specific methods and model level required fields checking
 module Bulkrax
   module CsvParserDecorator
     include OerCsvParser
-    # @return [Array<String>]
-    def required_elements
-      if Bulkrax.fill_in_blank_source_identifiers
-        %w[title resource_type]
-      else
-        ['title', 'resource_type', source_identifier]
-      end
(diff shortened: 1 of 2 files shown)


app/parsers/bulkrax/csv_parser_decorator.rb
```diff
@@ -1,41 +1,65 @@
...
+          # checks the field itself
+          # any parser_mappings fields terms from `config/initializers/bulkrax.rb`
+          # or any keys that has sequential numbers like creator_1
+          (record[field] ||
+            mapped_from(field).map { |f| record[f] }.first ||
+            handle_keys_with_numbers(field, record)).blank?
+        end
+      end
+
+      def determine_required_fields_for(model)
```
jeremyf, line 37:
```suggestion
      def determine_required_fields_for(model)
      # TODO: Revisit when Valkyrized as we can use Hyrax::ModelRegistry
```

jeremyf (approved):

Issue #973, opened 2024-02-20
Dates not sorting full string


ndroark:
In some cases, dates in the Date Created field fomatted as yyyy-mm-dd are only sorting by part of the string. Example here: https://wabash.hykucommons.org/collections/d4efac9d-24d2-472e-85cd-1c7da1319ccd?utf8=%E2%9C%93&sort=date_ssi+desc&per_page=10&locale=en

Relevant Slack thread: https://assaydepot.slack.com/archives/C0313NKC08L/p1708455332886429

jillpe:
SoftServ QA:

It did sort the works correctly via the date created, however, the date created is not visible on the works on the public collection show page.

[Collection created for QA](https://dev.commons-archive.org/collections/32d2f254-abad-48c8-a7e6-845ff5f7351a?utf8=%E2%9C%93&sort=date_ssi+asc&per_page=10&locale=en&view=list)



![Image](https://github.com/scientist-softserv/palni-palci/assets/84697174/043d10ed-a4a0-416c-b042-28c549c16571)

laritakr:
The reindex did not complete correctly. 

```
second = SolrDocument.find('1e392dac-37de-4d57-aa40-83020bbae320')
irb(main):008:0> second['date_ssi']
=> "2001"

GenericWork.find(second.id).update_index
irb(main):013:0> second['date_ssi']
=> "2001-01-18"
```

ShanaLMoore:
We reindex the collection of concern and see the issue got resolved so perhaps there were some issues when we reindexed before. We will need to run another reindex over the weekend.

PR #985, opened 2024-03-07
🐛 Fix date indexing to return more than YYYY

kirkkwang:
# Story

Prior to this commit when the date_created property was indexed, only the year was indexed.  If the input was something like "2024-01-01" then `date_ssi` would only be "2024".  This commit will index YYYY-MM-DD, YYYY-MM, and YYYY if given.  Anything other than those three formats will be indexed as is and won't work correctly with sorting or range queries.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/973

YYYY-MM-DD, YYYY-MM, and YYYY dates were indexed just as YYYY.

YYYY-MM-DD, YYYY-MM, and YYYY dates are indexes as YYYY-MM-DD, YYYY-MM, and YYYY respectively.

### DIFF
--- a/app/indexers/app_indexer.rb
+++ b/app/indexers/app_indexer.rb
@@ -30,12 +30,23 @@ class AppIndexer < Hyrax::WorkIndexer
   end
 
   def add_date(solr_doc)
-    # The allowed date formats are either YYYY, YYYY-MM, or YYYY-MM-DD
-    # the date must be formatted as a 4 digit year in order to be sorted.
-    valid_date_formats = /\A(\d{4})(?:-\d{2}(?:-\d{2})?)?\z/
     date_string = solr_doc['date_created_tesim']&.first
-    year = date_string&.match(valid_date_formats)&.captures&.first
-    solr_doc['date_tesi'] = year if year
-    solr_doc['date_ssi'] = year if year
+    return unless date_string
+
+    date_string = pad_date_with_zero(date_string) if date_string.include?('-')
+
+    # The allowed date formats are either YYYY, YYYY-MM, or YYYY-MM-DD
+    valid_date_formats = /\A(\d{4}(?:-\d{2}(?:-\d{2})?)?)\z/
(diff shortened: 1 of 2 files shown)


spec/indexers/app_indexer_spec.rb
```diff
@@ -24,4 +24,34 @@
       expect(solr_document.fetch("account_cname_tesim")).to eq(account.cname)
     end
   end
+
+  describe "#generate_solr_document" do
+    context "when given a date with a YYYY-MM-DD format" do
```
jeremyf, line 29: What about "2024-1-12"? Or "2024-13-50" or "2024-12-24-0"?

I think we should coerce "2024-1-12" to "2024-01-12" (using `sprintf` perhaps)?

jeremyf (changes requested):
See comments.

app/indexers/app_indexer.rb
```diff
@@ -31,11 +31,14 @@ def full_text(file_set_id)
 
   def add_date(solr_doc)
     # The allowed date formats are either YYYY, YYYY-MM, or YYYY-MM-DD
-    # the date must be formatted as a 4 digit year in order to be sorted.
-    valid_date_formats = /\A(\d{4})(?:-\d{2}(?:-\d{2})?)?\z/
+    valid_date_formats = /\A(\d{4}(?:-\d{2}(?:-\d{2})?)?)\z/
     date_string = solr_doc['date_created_tesim']&.first
```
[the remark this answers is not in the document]
kirkkwang: ah yes good point

spec/indexers/app_indexer_spec.rb
kirkkwang: Hmmm, I suppose we could, but the help text clearly says what date formats are allowed
https://github.com/scientist-softserv/palni-palci/blob/i973-date-sorting/config/locales/hyrax.en.yml#L1142

jeremyf (approved):

Issue #448, opened 2023-04-25
Remove Custom Mediated Deposit Workflow from options for Workflows on admin sets


crisr15:
# Summary

Currently there are two options for mediated deposit in Palni/Palci, but both function the same.  This is confusing for users and they should only have one options.  Remove the Custom Mediated Depot Workflow for Hyku Commons from the options and leave only the one step mediated deposit workflow and default workflow.  

# Acceptance Criteria

- [x] Write a script to migrate all Admin sets on the Custom Mediated Deposit Workflow for Hyku Commons to the One-Step mediated deposit workflow.  This script needs to run across all tenants.  
- [x] Change the default for an admin set to be "One-step mediated deposit workflow"
- [x] Remove the Custom Mediated Deposit Workflow for Hyku Commons from the options on the Admin Set Workflow Page

# Testing Instructions
- check that admin sets previusly set to teh Custom Mediated Deposit Workflow for Hyku Commons are now set to One-Step Mediated deposit.  Sample admin set in staging: https://dev.commons-archive.org/admin/admin_sets/b5c672a7-353e-42cf-a63e-b214cfbea9e5?locale=en
- Create a new Admin set.  Confirm that the workflow is automatically set to One-Step Mediated deposit workflow. 
- On the workflow page of your new admin set confirm that there is not an option to pick the Custom Mediated Deposit Workflow for hyku commons.  

# Notes

aprilrieger:
The script has already been run on staging and production to reassign the workflow 'hyku_commons_mediated_deposit' to 'one_step_mediated_deposit'.

labradford:
Client reports that new works that are deposited into the new default workflow are not showing up on the Review Submissions tab

labradford:
Newly deposited works not being added to the approval queue. Newly deposited work (https://nic2.commons-archive.org/concern/generic_works/dccec165-2513-4dd8-8213-b8bc1887c4aa?locale=en) shows "pending approval" status but the approval ribbon doesn't appear. Admin set currently using "one-step mediated deposit," though the "custom" workflow still appears as an option.

labradford:
from Nic

> In some repositories this change is working, in others (https://au-archives.hykucommons.org/admin/workflows?locale=en) I submitted a test work in an admin set with mediated deposit, and the queue shows "no data available in table" and depositors in the already published tab are still pulling from creator. Possibly related to https://github.com/scientist-softserv/palni-palci/issues/448 ?

labradford:
[...]

PR #994, opened 2024-03-11
🧹 Hide custom Hyku Commons workflow

kirkkwang:
# Story

This commit will hide the custom Hyku Commons workflow from the admin workflow form.  This is a temporary measure to ensure that the option is not going to show up in the admin interface to reduce confusion.  The long-term fix will be to create a script and delete the custom workflow from the database from each tenant, however it was warned that working with Siptiy::Workflow might be a little tricky.

Ref:
  - https://github.com/scientist-softserv/palni-palci/issues/448

# Screenshots / Video

image

image

### DIFF
--- a/app/views/hyrax/admin/admin_sets/_form_workflow.erb
+++ b/app/views/hyrax/admin/admin_sets/_form_workflow.erb
@@ -1,0 +1,30 @@
+<%# OVERRIDE Hyrax v3.6.0 to hide custom mediated workflow for Hyku Commons %>
+
+<div id="workflow" class="tab-pane">
+  <div class="panel panel-default labels">
+    <%= simple_form_for collection_permission_template_form_for(form: @form),
+      url: [hyrax, :admin, @form, :permission_template],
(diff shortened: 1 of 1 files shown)


laritakr (changes requested):
I don't think we want to override the view and leave the data behind. Doing this half-way adds more complication for the knapsack work. So unless someone disagrees, I believe we should leave this until we can do the task fully.

I think we want to document the data migrations needed:  Before deleting the workflow, you need to update Sipity::Entity to use the workflow_id that is being substituted for the one being deleted. This is a database table, so the command to update would be fairly simple... something like `Sipity::Entity.where(workflow_id: 3).update_all(workflow_id: 1)` When you delete a workflow, you need to delete all of the child records that are unused. 

See https://github.com/samvera/hyrax/blob/main/artifacts/entity-relationship-diagram.pdf for the data diagram. The workflow definitions are the black tables. Workflow_id would be in the WorkflowRole, WorkflowState, and WorkflowAction tables. Lower level entries could also be affected as long as the entries aren't used with a different workflow. It's not ideal to have any orphaned definition records.

PR #995, opened 2024-03-11
🐛 Account for when result is not a string

kirkkwang:
This commit will fix a bug where if the csv has a `location` header but no value.


laritakr (approved):
Making another ticket to address the flaky specs.

Issue #1042, opened 2024-06-29
Re-allow CSVs without model column to fallback on default work type


bkiahstroud:
# Story

Importing a CSV without a model column leads to the importer failing to run and logging this error: `NameError - uninitialized constant Hyrax::Form Did you mean? Hyrax::Forms`. 

CSVs imported without a model column should fallback on `Bulkrax.default_work_type`. 

# Acceptance Criteria

- [ ] Importing a CSV without a model column will successfully create `GenericWorks` instead of throwing an error 

# Testing Instructions and Sample Files

1. Login as an admin 
2. Navigate to Dashboard > Importers > New Importer 
3. Upload a valid CSV that does not have a model column 
4. Click Create and Import 
5. Validate that the importer successfully creates GenericWorks instead of failing with an error 

# Notes

<details><summary>Stack trace</summary>

```
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/inflector/methods.rb:285:in `const_get'
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/inflector/methods.rb:285:in `block in constantize'
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/inflector/methods.rb:281:in `each'
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/inflector/methods.rb:281:in `inject'
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/inflector/methods.rb:281:in `constantize'
/usr/local/bundle/gems/activesupport-5.2.8.1/lib/active_support/core_ext/string/inflections.rb:68:in `constantize'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:42:in `determine_required_fields_for'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:26:in `missing_fields_for'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:11:in `block in valid_import?'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:9:in `each'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:9:in `each_with_object'
/app/samvera/hyrax-webapp/app/parsers/bulkrax/csv_parser_decorator.rb:9:in `valid_import?'
[...]

PR #1043, opened 2024-06-29
i1042 Fall back on default work type when model is empty

bkiahstroud:
# Story

Ref 
- #1042 

Importing a CSV without a model column leads to the importer failing to run and logging this error: `NameError - uninitialized constant Hyrax::Form Did you mean? Hyrax::Forms`. 

CSVs imported without a model column should fallback on `Bulkrax.default_work_type`.


kirkkwang (approved):
LGTM!

Issue #1063, opened 2024-12-09
Tenant storage list


jillpe:

jillpe:
https://assaydepot.slack.com/archives/C0313NKC08L/p1733767903860929

jillpe:
https://github.com/scientist-softserv/palni-palci/blob/main/lib/tasks/tenants.rake

aprilrieger:
```
uindy.hykucommons.org: 632.8499999999998 Total MB / 482 Works
au-archives.hykucommons.org: 21165.829999999998 Total MB / 624 Works
ambs.hykucommons.org: 1396.7000000000005 Total MB / 144 Works
palsave.hykucommons.org: 2503.750000000001 Total MB / 77 Works
grace.hykucommons.org: 45610.78999999996 Total MB / 2505 Works
palni.hykucommons.org: 217.48000000000002 Total MB / 56 Works
hanover.hykucommons.org: 30672.939999999908 Total MB / 3247 Works
cairn.hykucommons.org: 1082.0099999999998 Total MB / 123 Works
allegheny.hykucommons.org: 0 Total MB / 0 Works
juniata.hykucommons.org: 5284.530000000002 Total MB / 109 Works
wvu.hykucommons.org: 95.81 Total MB / 2 Works
goshen.hykucommons.org: 10235.26 Total MB / 1911 Works
palci-demo.hykucommons.org: 714.58 Total MB / 89 Works
alpa.hykucommons.org: 1423.7299999999923 Total MB / 372 Works
chatham.hykucommons.org: 35.01 Total MB / 1 Works
viva-test.hykucommons.org: 71.13000000000001 Total MB / 16 Works
marian.hykucommons.org: 9192.449999999943 Total MB / 2520 Works
sma-ir-library.hykucommons.org: 4561.140000000005 Total MB / 1621 Works
earlham.hykucommons.org: 23856.700000000095 Total MB / 5904 Works
saint-francis.hykucommons.org: 215.44999999999996 Total MB / 37 Works
louis-test.hykucommons.org: 5.12 Total MB / 11 Works
dickinson.hykucommons.org: 4160.870000000007 Total MB / 2456 Works
huntington.hykucommons.org: 7978.350000000004 Total MB / 278 Works
franklin.hykucommons.org: 2638.319999999999 Total MB / 286 Works
saint-marys.hykucommons.org: 314.74000000000007 Total MB / 57 Works
[...]

PR #1064, opened 2024-12-10
i1063-fix-tenants-rake

aprilrieger:
#1063 

Updates the tenants.rake to:
- turn off sentry reporting during the task and then re-enables at the end
- add graceful error handling in order to get the tenants data for client

### DIFF
--- a/lib/tasks/tenants.rake
+++ b/lib/tasks/tenants.rake
@@ -5,7 +5,7 @@ namespace :tenants do
   task calculate_usage: :environment do
     @results = []
     Account.where(search_only: false).find_each do |account|
-      next unless account.cname.present? # Skip accounts without a cname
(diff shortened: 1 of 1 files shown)


lib/tasks/tenants.rake
```diff
@@ -1,55 +1,72 @@
 # frozen_string_literal: true
 
+# Temporarily disable Sentry for this task
+Raven.configure do |config|
+  config.silence_ready = true
+end
```
bkiahstroud, line 6: Will this disable Raven for the running application while the task is running?

aprilrieger: Good point, I believe so.

bkiahstroud (approved):

